PR #3, opened 2022-07-28
Don't use a makefile, run tasks from `npm-scripts`

sndrs:
this bugged me all last night.

is this just massive over-engineering? are the reasons we [added it in frontend](https://github.com/guardian/frontend/pull/11022) and [then to DCR](https://github.com/guardian/dotcom-rendering/pull/28) no longer valid? 

e.g. does every dev now expect to use a node version manager? maybe...


.github/workflows/ci.yml
```diff
@@ -10,8 +10,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - name: Setup Node
-        uses: actions/setup-node@v3
+      - uses: pnpm/action-setup@v2.0.1
```
tjsilver, line 13: Looks like it's failing because it can't find a version - does it need something like this? 
`uses: pnpm/action-setup@v2.2.2
        with:
          version: 6.0.2`

tjsilver: Also should we use the latest version, i.e. v2.2.2?

sndrs: ah yeah, i think later version will use the `packageManager` key in the `package.json`. lazy copying :)

sndrs:
closing for now. i think we think we _should_ go with makefile approach for now. at least this PR lists the areas to address in the future

PR #6, opened 2022-07-28
add `@guardian/tsconfig`

sndrs:
Add `@guardian/tsconfig`.

It has no build step so currently changesets/nx does not need any special config (due to changesets not expecting Nx's default behaviour of building everything to a root-level `dist`)


tjsilver (approved):
Exciting!

PR #7, opened 2022-07-29
Do not mandate `fnm`

sndrs:
just check the current Node version and prompt if it's wrong (let the user choose how to manage their own Nodes)


tjsilver (approved):
This is good, because asdf is being recommended for the M1 machines.

joecowton1 (approved):

sndrs:
>This is good, because asdf is being recommended for the M1 machines.

really? that's interesting. it's a rubbish DX for node (compared to fnm)

PR #9, opened 2022-08-02
Fix issues caused by TS upgrade

joecowton1:
## What does this change?

Fixes issues caused by upgrading TS to `4.7.4`.

We should probably implement these changes in the original `libs` repo before porting the code over, however for the purposes of the NX spike, this allows us to succesfully build libs within the new monorepo.


libs/libs/jest.config.ts
```diff
@@ -1,13 +1,14 @@
 /* eslint-disable */
 export default {
 	displayName: 'libs',
+	clearMocks: true,
```
joecowton1, line 4: This comes from the current libs test config.

libs/libs/scripts/update-readme.ts
```diff
@@ -8,7 +8,8 @@
 
 import fs from 'fs';
 import path from 'path';
-import pkg from '../package.json';
+// TODO: import this from root package
+// import pkg from '../../../package.json';
```
joecowton1, line 12:
```
File '/Users/Joe_Cowton/code/csnx/package.json' is not under 'rootDir' 'libs/libs'. 'rootDir' is expected to contain all source files.
```

libs/libs/src/cookies/getCookie.ts
```diff
@@ -22,9 +22,11 @@ export const getCookie = ({
 
 	if (cookieVal.length > 0) {
 		if (shouldMemoize) {
-			memoizedCookies.set(name, cookieVal[0]);
+			// TODO: extract this to a function
+			memoizedCookies.set(name, cookieVal[0] ? cookieVal[0] : '');
```
joecowton1, line 26: Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
  Type 'undefined' is not assignable to type 'string'.

libs/libs/src/coreWebVitals/index.test.ts
```diff
@@ -60,7 +60,7 @@ const mockConsoleWarn = jest
 
 const spyLog = jest.spyOn(logger, 'log');
 
-const setVisibilityState = (value: VisibilityState = 'visible') => {
+const setVisibilityState = (value: DocumentVisibilityState = 'visible') => {
```
joecowton1, line 63: Cannot find name 'VisibilityState'.ts(2304)

libs/libs/src/datetime/timeAgo.ts
```diff
@@ -1,6 +1,6 @@
 type Unit = 's' | 'm' | 'h' | 'd';
 
-const shortMonth = (month: number): string =>
+const shortMonth = (month: number) =>
```
joecowton1, line 3:
```
Type 'string | undefined' is not assignable to type 'string'.
Type 'undefined' is not assignable to type 'string'.
```
Type removed as TS can auto infer it.

libs/libs/src/storage/storage.test.ts
```diff
@@ -16,8 +16,8 @@ describe.each([
 	let setSpy: jest.SpyInstance;
 
 	beforeEach(() => {
-		getSpy = jest.spyOn(native.__proto__, 'getItem');
-		setSpy = jest.spyOn(native.__proto__, 'setItem');
+		getSpy = jest.spyOn(native['__proto__'], 'getItem');
```
joecowton1, line 19:
```
Property '__proto__' comes from an index signature, so it must be accessed with ['__proto__'].ts(4111)
```

libs/libs/tsconfig.lib.json
```diff
@@ -2,8 +2,7 @@
 	"extends": "./tsconfig.json",
 	"compilerOptions": {
 		"outDir": "../../dist/out-tsc",
-		"declaration": true,
-		"types": []
```
joecowton1, line 6: Removing this empty types array fixes the error with `@types/node`:
```
 Cannot find module 'fs' or its corresponding type declarations.
```

PR #30, opened 2022-08-08
Use @nrwl:web/rollup for build 

joecowton1:
## What does this change?

Uses the [@nrwl/web:rollup](https://nx.dev/packages/web/executors/rollup#@nrwl/web:rollup) executor to build the `tsconfig` package rather than `@nrwl/js:tsc`.

Also adds a `babel.config.json` to fix [this issue](https://github.com/nrwl/nx/issues/3184#issuecomment-674770770).

Also updates all `nrwl` packages to the latest version.

I had to hardcode the asset paths rather than using `*.md` - not sure why that is. 

## How to test

Pull down the project and run `make build`. The `dist` folder should contain both `index.js` and `index.cjs` along with the other project files.


babel.config.json
```diff
@@ -0,0 +1,4 @@
+{
```
tjsilver, line 1: Not sure what this is for - I tried running `make build` both before and after deleting it and the outcome looked the same.

joecowton1: That might just be because nx has cached it: if you run the following it should throw an error with the babel config:
```
pnpm nx run-many --target=build --all=true --skip-nx-cache

```

joecowton1: The error is: 
```
 Error during bundle: Error: Babel was run with rootMode:"upward" but a root could not be found when searching upward from "/Users/Joe_Cowton/code/csnx/libs/tsconfig".
 ```
 
 I linked the issue in the original PR comment but for reference it's here: https://github.com/nrwl/nx/issues/3184#issuecomment-674770770

File: libs/tsconfig/project.json [hunk 2]
```diff
@@ -1,16 +1,22 @@
...
-				"main": "libs/tsconfig/src/index.ts",
 				"tsConfig": "libs/tsconfig/tsconfig.lib.json",
-				"assets": ["libs/tsconfig/*.md", "libs/tsconfig/tsconfig.json"]
+				"assets": [
+					"libs/tsconfig/readme.md",
+					"libs/tsconfig/CHANGELOG.md",
+					"libs/tsconfig/tsconfig.json"
+				],
+				"format": ["esm", "cjs"]
 			}
```
tjsilver, line 20: I see that the rollup executor uses babel by default, but can accept either babel or swc as compiler - we could change it to swc with the option `"compiler": "swc"` and that should be faster - and would presumably negate the need for the babel config file.

joecowton1: Started looking at that earlier - there are advantages (build time) and disadvantages (doesn't export types) so we'd need to weigh those up: https://github.com/guardian/csnx/pull/32

joecowton1: I don't think having a babel config file is inherently a bad thing though unless you have a specific reason for thinking that?

babel.config.json
tjsilver: Ah yes, got that error too. It was solved when I changed the compiler to swc.

PR #33, opened 2022-08-11
Spike package executor

tjsilver:
## What does this change?

1. Adds a custom executor `@csnx/npm-package:build` that uses Rollup to build and bundle npm packages.
2. Uses it to build `@guardian/tsconfig`.
3. Updates decision record to reflect how we got to this decision, and the alternatives that we tried.


package.json
```diff
@@ -12,18 +12,21 @@
...
 		"@nrwl/eslint-plugin-nx": "14.4.3",
-		"@nrwl/jest": "14.4.3",
+		"@nrwl/jest": "14.5.4",
 		"@nrwl/js": "14.4.3",
 		"@nrwl/linter": "14.4.3",
 		"@nrwl/node": "^14.4.3",
 		"@nrwl/nx-cloud": "latest",
 		"@nrwl/workspace": "^14.4.3",
+		"@swc-node/register": "^1.5.1",
+		"@swc/core": "^1.2.224",
```
tjsilver, line 29: NX uses these instead of `ts-node` when it compiles internal builders and executors on the fly, for speed.

PR #35, opened 2022-08-11
WIP: Add eslint and eslint-typescript configs

joecowton1:
**Still needs doing:** add test file and get tests and linting wired up with nx / jest config

## What does this change?

Adds the existing `@guardian/eslint-config` and `@guardian/eslint-config-typescript` config packages to the csnx monorepo.

To do this we have:
- used nx to generate a new `libs` project using the name `prettier`
- gutted the generated tsconfig files to remove unused rules
- removed generated test files
- replaced the generated package.json with the existing prettier package.json 
- replaced the default executor in project.json with `@csnx/npm-package:build`
- updated the prettier dependency to use: `"@guardian/package": "workspace:*",`
- updated the readme

**Note:** running `make build` is currently outputting `esm` and `cjs` folders correctly, however within those folders we have multiple index files: (`index.js` and `index2.js`, where `index2.js` is imported by `index.js`). 

Maybe this is fine? If this is not fine the following seem to resolve it:
- set `preserveModules` to false in the rollup config
- make the file `index.ts`
- use `export default` rather than `module.exports` in `src/index.js`

## How to test

run `make build`

do a release via changesets

consume the lib


sndrs:
I've made some update to this:

- use the updated `@csnx/npm-package:build`
- remove our internal usage of these configs
  - will re-add that in a separate PR
- removed the changeset until they're working internally

PR #42, opened 2022-08-12
Do not apply prettier formatting as part of linting

sndrs:
## What does this change?

- removes checking for `prettier` conformance from our ESLint configs
- move some config from the TS config up to the standard config

### Before

We extended `plugin:prettier/recommended` which

1. used `eslint-config-prettier` to disable any white-space formatting rules that would conflict with our prettier config
2. used `eslint-plugin-prettier` to lint for formatting errors that did not match our prettier config

This is quite expensive, and although it means you could use `--fix` to apply prettier, it's not as fast as using prettier directly.

### After

We still use `eslint-config-prettier` to avoid conflicts with our `prettier` config, but we no longer lint for errors (and therefore also don't fix them).


mxdvl (approved):
I have found that linting can become unbearably slow at times and have see conflicts between ESLint and Prettier causing irreconciliable linting loops, so this is a much welcome simplification.

PR #45, opened 2022-08-15
Dont warn about missing peerdeps for these packages

sndrs:
they’re installed in the root, no one else needs to worry


tjsilver (commented):
👍

PR #59, opened 2022-08-24
add `@guardian/libs`

sndrs:
- adds `@guardian/libs`
- converts `lint-staged` config to use config files, so we can custom configs for subfolders


sndrs:
closing this in favour of a more, smaller changes

PR #64, opened 2022-08-31
Improve node version checking

sndrs:
- ~~if the current node version is wrong, tries to use a locally installed `asdf`, `fnm` or `nvm` (in that order) to switch to the correct version, installing it if necessary.~~
- ~~tries to initialise a local version manager if one is installed but hasn't been initialised (e.g. we're in a non-interactive shell like vscode's git gui, github desktop, etc)~~
- ~~errors with a prompt to install `fnm` if we can't do anything~~

[updated]

- still performs a simple check, per @akash1810's suggestion, but tries to offer help about what to do.

I think (hope) this a more maintainable option while still prioritising devex. the thinking being that it's useful to be able to prep you env to handle these kind of requirements, but you may also no know you've kind of done that already (e.g. installed a node version manager for something else).

hopefully it will also improve the situation when committing. i'm still not 100% sure about whether it's better to prep the env for people or not. but it feels like the last version was a bit of a hodgepodge way of doing it. either we go full [`nve`](https://www.npmjs.com/package/nve) type approach, or just do this lightweight check.


tools/scripts/use-node-version
```diff
@@ -0,0 +1,158 @@
...
+		# this will also be false if no node is installed
+		return 1
+	fi
+}
+
+# try to load the correct version of node using a local node version manager (if
+# there is one)
+load_correct_node() {
+	# try using asdf
+	if asdf_installed; then
```
ob6160, line 45: is the priority of this chain determined by an official CSTI recommendation? it'd be nice to be able to allow users to choose which tool they use instead of choosing for them, say if i had both `asdf` and `fnm` installed..

or at the very we could give them the option of cancelling the process before automatically going ahead and installing with a given tool

sndrs: it's based on the recommendations for the new laptops.

if you do have asdf _and_ fnm installed, you're going to be in a bit of a pickle anyway (i'm guessing asdf's shims will win out?) but if you _have_ got both working, what's the value in us asking which one you want to use?

the intention was to be as unobstructive as possible – just get hold of the right version of node in the way it looks like you're set up to things

tools/scripts/use-node-version
```diff
@@ -0,0 +1,158 @@
...
+		fi
+		return 0
+	fi
+
+	# can't find fnm
+	return 1
+}
+
+install_with_fnm() {
+	fnm use --install-if-missing
```
ob6160, line 119: We have comments in the equiv method for `asdf` and `nvm` — should this one have a comment too?

tools/scripts/use-node-version
```diff
@@ -0,0 +1,158 @@
...
+install_with_nvm() {
+	# check if nvm has already installed the version we need taken from
+	# https://github.com/nvm-sh/nvm#bash
+	existing_installation=$(nvm ls --no-colors "$nvmrc_version" | tail -1 | tr -d '\->*' | tr -d '[:space:]')
+
+	if [[ "$existing_installation" == "N/A" ]]; then
+		echo -e "${dim}Installing Node v$nvmrc_version with ${blue}nvm${reset}"
+		nvm install "$nvmrc_version"
+	else
+		# echo -e "Using Node ${blue}$(node -v)${reset}"
```
ob6160, line 147: can this line be removed?

tools/scripts/use-node-version
ob6160: good point around having both installed — hadn't considered that might be quite a rare scenario.

what did you think about having some kind of user confirmation prompt?

tools/scripts/use-node-version
```diff
@@ -0,0 +1,158 @@
+#!/usr/bin/env bash
+
+# Check whether the current node version matches the .nvmrc version.
```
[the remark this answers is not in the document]
sndrs: it's somewhat outside the scope of this PR, but i don' think we can cater for that scenario. i'm not sure if that would present a real problem thoguh?

tools/scripts/use-node-version
```diff
@@ -0,0 +1,158 @@
...
+	else
+		# echo -e "Using Node ${blue}$(node -v)${reset}"
+		nvm use "$nvmrc_version"
+	fi
+}
+
+using_correct_node || load_correct_node || {
+	# we've run out of options so just warn the user and exit
+	echo -e "${red}Could not find the required Node version (v$nvmrc_version)${reset}"
+	echo -e "${dim}Consider using ${bold}fnm${reset}${dim} to manage Node versions:${reset} ${blue}https://github.com/Schniz/fnm#installation${reset}"
```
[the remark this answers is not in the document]
sndrs: i went with fnm because if you don't have _any_ way to manage node already, it seems the simplest to install and live with. asdf is more for power users, and nvm is slower.

tools/scripts/use-node-version
sndrs: what would we prompt for?

tools/scripts/use-node-version
sndrs: do you mean an echo to the terminal?

tools/scripts/use-node-version
ob6160: i was thinking, prompt to confirm that they want to continue with the installation — but i just remembered that fnm does that anyway:
Screenshot 2022-08-31 at 18 36 53

tools/scripts/use-node-version
ob6160: yep — used the wrong word :)

tools/scripts/use-node-version
sndrs: i'm not sure - is there any value in saying no? i woudn't want to change anything global on the user's system (e.g. setting the system default node) but here we're just using a node version manager they've already installed themselves to install a version of node, i think that's ok?

the intention here is to make the fact that they _need_ to install it a non-issue (in a non-invasive way)

tools/scripts/use-node-version
sndrs: got you - no fnm is very explicit here. the other two methods are a less so, they get the extra info

tools/scripts/use-node-version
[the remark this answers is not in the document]
ob6160: I was wondering if a newer version of node is used to install packages for projects that specify an older version of node whether that could cause problems due to breaking API changes that might happen between versions. I'm only raising this because I've sometimes seen projects break when using a newer version of node to install deps (this was back when the platform was a lot younger to be fair).

I suppose another consideration might be that — if/when we move other projects to the monorepo — developers might want to develop using the version of node that their project targets when deployed to production.

You're right though, this is outside the scope of this PR and is probably a discussion to have elsewhere

tools/scripts/use-node-version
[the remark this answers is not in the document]
ob6160: agreed :)

tools/scripts/use-node-version
ob6160: I see - they do a great job with their logging

tools/scripts/use-node-version
ob6160: it seems to be a fairly standard pattern where if program is about to run a command that'll download a new executable to a computer it gives a choice to say no before doing so (like you get with apt/brew/pacman etc)

akash1810:
This adds a lot of complexity, and assumes the tools being used to manage node versions. I'd suggest keeping the logic in the repo super simple, and make it the user's responsibility to use the right version, however they choose to.

```bash
currentNodeVersion=$(node -v)
expectedNodeVersion=$(cat .nvmrc)

if [ "$currentNodeVersion" != "$expectedNodeVersion" ]; then
  echo "Using wrong version of Node. Required ${expectedNodeVersion}. Running ${currentNodeVersion}."
  exit 1
fi

// we can now guarantee the node version is correct
```

sndrs:
> I'd suggest keeping the logic in the repo super simple

yeah i'm torn on this. that's how this file started life :)

i think there's an awful lot of value in making the whole thing as simple to run as possible. 

for example, originally you could clone DCR, run `make dev` and you were working.

this is trying to follow that model, but DCR also did things like install yarn etc, which I think is a step too far (it's changing the system globally).

this is an attempt to use the tools you may have already installed (that is, follow a pattern the user has already chosen), and to automate the annoying things you have to do with those tools (install a missing node, remember to run them etc).

> and make it the user's responsibility to use the right version, however they choose to.

i _think_ i think that the responsibility to make it safe to run the code in the repo should be maintainer's? at least, this change is is obviously strongly opinionated in that direction.

(i should point out i tried to make it as fast as possible to continue if the user _has_ already made sure they're using the correct version).

i'd guess mode devs would expect to accommodate an nvmrc file, so much of this code may never run. the people most likely to need it (i'm guessing) would those least likely to be familiar with setting up node etc. 

it's an attempt to make it as trivial as possible for them to work in the repo, rather than needing to understand an ecosystem before they make changes.

that's at least the case _for_ this approach...

> This [...] assumes the tools being used to manage node versions

it only tries to lend a hand for these three, that's true, but if you're using `n` or something, you're in the situation you'd be in if we just exited with a warning. it doesn't mandate using one of them

tools/scripts/use-node-version
sndrs: > if program is about to run a command that'll download a new executable to a computer it gives a choice to say no before doing so (like you get with apt/brew/pacman etc)

is this true? almost everything you download from brew or npm etc comes with dependencies that are installed without prompting, because you've elected to install the thing that depends on them.

if you try to run a task that has a dependency, and we download that dependency – with a notification that that's what we're doing and using the tool you've installed to do that – is that different?

ob6160: > because you've elected to install the thing that depends on them

I'd argue that in my experience it's quite rare that just running `yarn` installs a specific version of node on my computer on top of installing the project dependencies. So the extra invocation of `fnm`,`asdf`,etc.... doesn't fall within the remit i gave my `yarn install` when I elected to install my dependencies.

sndrs:
for some extra context https://mail.google.com/chat/u/0/#chat/space/AAAAWwBdSMs/LwcXpTnEG1Q

oliverlloyd:
Well, I just cloned and tried to open my first PR and got hung up on the very issue that this PR would resolve so I think this added complexity is worth the improved developer experience it offers. +1 from me

ob6160 (approved):
🌟

tjsilver (approved):

PR #86, opened 2022-09-09
rename `tools/plugins` to `tools/nx-plugins`

sndrs:
## What are you changing?

- rename `tools/plugins` to `tools/nx-plugins`
- flatten files `tools/nx-plugins/npm-package`
- stop building `@csnx` packages

## Why?

- makes the tools file structure easy to understand
- there's no point in building an internal package


tjsilver (approved):
This looks like a sensible refactor to make things clearer.

PR #88, opened 2022-09-09
Use rollup's output to set the entry fields in `package.json`

sndrs:
## What are you changing?

- use rollup's output info the set the paths for the `main`, `module` and `exports` fields

## Why?

- the `main`, `module` and `exports` field paths were hardcoded to the module format path, but this path is not always correct (e.g. if the code in `src` imports from somewhere outside of there)

### before 

```json
  "exports": "./esm/index.js",
  "main": "cjs/index.js",
  "module": "esm/index.js",
```

### after
```json
  "exports": "./esm/index.js",
  "main": "cjs/index.js",
  "module": "esm/index.js",
```
or
```json
  "exports": "./esm/src/index.js",
  "main": "cjs/src/index.js",
  "module": "esm/src/index.js",
```
or whatever rollup tells us it's built

### DIFF
--- a/libs/@guardian/libs/README.md
+++ b/libs/@guardian/libs/README.md
@@ -97,7 +97,7 @@ import { loadScript, storage, ...etc } from '@guardian/libs';
 
 ### TypeScript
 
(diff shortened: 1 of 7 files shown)


tjsilver (approved):
Looks great!

PR #101, opened 2022-09-15
Add skeletons for the `source` packages

sndrs:
## What are you changing?

- adds skeleton code around `@guardian/source-foundations`, `@guardian/source-react-components` and `@guardian/source-react-components-development-kitchen`
- doesn't include the actual source code, because that still needs to live in the main repo for now

## Why?

- prep for transferring from https://github.com/guardian/source


tjsilver (dismissed):
Great skeleton 👍

PR #119, opened 2022-10-05
create global csnx window type

sndrs:
## What are you changing?

- moving the window ambient type declaration from libs to the project root

## Why?

- makes it available for all projects to include and define props on


sndrs:
fyi @mxdvl @JamieB-gu

tjsilver (approved):
👍

PR #132, opened 2022-10-10
Add SpecialReportAlt theme article format

DanielCliftonGuardian:
Co-Authored-By: Ioanna Kokkini <ioannakok@users.noreply.github.com>

## What are you changing?

This adds `SpecialReportAlt` theme in `ArticleSpecial`.

**BREAKING CHANGE**: Consumers of `ArticleFormat` and/or `ArticleTheme` will need to update to handle this change. This may involve updating component props, fixtures, switch and if statements, etc.

## Why?
We need to support a new article type of `SpecialReportAlt`.

### DIFF
--- a/.changeset/dull-tools-knock.md
+++ b/.changeset/dull-tools-knock.md
@@ -1,0 +1,5 @@
+---
+'@guardian/libs': major
+---
+
+This adds SpecialReportAlt theme in article format. We need to support a new type of special report. Consumers need to handle the new theme.
--- a/libs/@guardian/libs/src/format/ArticleSpecial.ts
+++ b/libs/@guardian/libs/src/format/ArticleSpecial.ts
@@ -1,4 +1,5 @@
 export enum ArticleSpecial {
 	SpecialReport = 5,
 	Labs = 6,
+	SpecialReportAlt = 7,
 }


.changeset/dull-tools-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/libs': major
```
joecowton1, line 2: Is this a breaking change or just a feature?

joecowton1:
Left a comment to this effect in the code, but i'm interested in whether this actually is a breaking change: if we're just adding a new type of theme will that break anything? My gut feeling is that consumers should be fine to continue as they were without it effecting anything.

.changeset/dull-tools-knock.md
JamieB-gu: I think it's a breaking change because of exhaustiveness checking? Any code that previously checked all variations of `Theme` will now have an extra variation that it's not checking for. For example:

```ts
const themeToString = (theme: ArticleTheme): string => {
  switch (theme) {
    case ArticlePillar.News:
      return 'news';
    case ArticlePillar.Sport:
      return 'sport';
    case ArticleSpecial.Labs:
      return 'labs';
    ...
  }
}
```

There's no case for `ArticleSpecial.SpecialReportAlt`, so the compiler will generate an error. It has to because otherwise the new case wouldn't be matched, resulting in `undefined` being implicitly returned, which is not equal to `string`.

ioannakok:
> 

Thanks for your comment @joecowton1! From a quick look in the DCR codebase, it seems there are places where we switch over `format.theme` without having a default case, e.g.
https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/lib/decidePalette.ts#L202-L217

At the moment TS is not complaining but we have a PR to fix this: https://github.com/guardian/dotcom-rendering/pull/6105 hence why we think this is a breaking change.

JamieB-gu (approved):

.changeset/dull-tools-knock.md
ioannakok: Yes, exactly what @JamieB-gu says. We do have places in DCR where we switch over `format.theme` without having a default case, e.g. https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/lib/decidePalette.ts#L202-L217

At the moment TS is not complaining but we have a PR to fix this: https://github.com/guardian/dotcom-rendering/pull/6105.

joecowton1: Thanks for clarifying 👍

joecowton1 (approved):

jamesgorrie (approved):

oliverlloyd:
It's true that this change could cause some consumers to break, so it is breaking in this (important!) sense. Arguably though, we should probably look to update this code to be more resilient and allow these sorts of changes to go out as features, which perhaps, semantically, they are?

PR #165, opened 2022-11-02
Publish storybooks to S3 with riffraff

sndrs:
just for a demo


sndrs:
cleaner version of #160

nicl:
@sndrs and @joecowton1 I've updated this now and it works. So should be 👍 to merge I think.

nicl:
Closing for now as going with a different approach, but we'l revisit easier ways to support static sites going forward!

PR #220, opened 2022-11-22
Add initial storybook config for Source

ob6160:
## What are you changing?

Adds storybook config for the following Source packages:
- `source-foundations`
- `source-react-components`
- `source-react-components-development-kitchen`

## Why?

- to allow storybook to build for each package

## Notes

Currently, three different storybooks are created with their own URL. We need to discuss how best to consolidate these as a 'Source' storybook, so that we can [link to it as before](https://guardian.github.io/source/?path=/story/source--page)

The addition of `backgrounds.ts` reinstates background colour mappings for the storybook, so the backgrounds appear inverted in the Storybook diff. Do not fear -  is actually flipping to the expected behaviour now the bg colour definitions exist.

We also rename `manager.js` to `manager.ts` so it is captured by the csnx eslint configured and the import of `theme.ts` is resolved correctly.


libs/@guardian/source-react-components-development-kitchen/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
 	"extends": "../../../tsconfig.base.json",
 	"files": [],
-	"include": [".", "../../../@types/window.d.ts"],
+	"include": [".", "../../../@types/window.d.ts", ".storybook/*"],
```
joecowton1, line 4: Do we need to add this here?

joecowton1 (approved):

libs/@guardian/source-react-components-development-kitchen/tsconfig.json
ob6160: without these, the hidden ".storybook" directory won't be captured by tsconfig as it'll ignore hidden files/dirs by default afaik

libs/@guardian/source-react-components/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
 	"extends": "../../../tsconfig.base.json",
 	"files": [],
-	"include": ["."],
+	"include": [".", ".storybook/*"],
```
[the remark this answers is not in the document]
ob6160: https://github.com/guardian/csnx/pull/220#discussion_r1029635099

libs/@guardian/source-react-components-development-kitchen/tsconfig.json
joecowton1: yeah i get that, i just mean if we weren't checking the storybook config files previously, do we need to start doing it now?

joecowton1: Is there any actual types to check within the config?

ob6160: ah good point - I think it's worth doing type checking against them.. can't think of a reason why we wouldn't. We also have types here: https://github.com/guardian/csnx/pull/220/files#diff-a03d3fafe3aae070b0e2c3e77cfb2f949ad227868d805467765a6299df112636R45 that @tjsilver added in one of our earlier preparation PRs to move Source in

joecowton1: ok if you think it makes sense then sure, just good to question the motivations if we're changing things 👍

ob6160:
> One for a separate PR but can we update the individual story titles so we don't have these nested titles? I don't think they're needed now each package has its own storybook.
> 
> ie rather than ` title: 'Packages/source-react-components/Accordion',`
> 
> Screenshot 2022-11-23 at 09 04 01
> 
> just use `title: Accordion`
> 
> Screenshot 2022-11-23 at 09 05 27

good spot, will note this down for a followup pr

PR #225, opened 2022-11-22
Use incremental compiler option by default

mxdvl:
## What are you changing?

- Enable the [incremental compiler option](https://www.typescriptlang.org/tsconfig#incremental) in tsc

## Why?

- Faster rebuilds


joecowton1 (approved):
Nice 👌

mxdvl:
Hopefull this work without issues… 🤞

PR #229, opened 2022-11-28
Remove stories for Source eslint plugins

ob6160:
## What are you changing?

- We're removing both Source eslint plugin README stories in favour of using their markdown equivalent.

## Why?

- We felt it was unnecessary to document this information as a story, especially because we don't render any components.
- This change would have required adding a new instance of AWS Amplify for each plugin.

### DIFF
--- a/libs/@guardian/eslint-plugin-source-foundations/README.stories.mdx
+++ b/libs/@guardian/eslint-plugin-source-foundations/README.stories.mdx
@@ -1,9 +1,0 @@
-import { Meta } from '@storybook/addon-docs';
-import ReadMe from './README.md';
-
-<Meta
-	title="packages/eslint-plugin-source-foundations/README"
-	parameters={{ previewTabs: { canvas: { hidden: true } } }}
-/>
-
-<ReadMe />
--- a/libs/@guardian/eslint-plugin-source-react-components/README.stories.mdx
+++ b/libs/@guardian/eslint-plugin-source-react-components/README.stories.mdx
@@ -1,9 +1,0 @@
-import { Meta } from '@storybook/addon-docs';
-import ReadMe from './README.md';
-
-<Meta
-	title="packages/eslint-plugin-source-react-components/README"
-	parameters={{ previewTabs: { canvas: { hidden: true } } }}
-/>
-
-<ReadMe />


joecowton1:
Just so I'm clear - do we still have versions of these in the storybook?

ob6160:
> Just so I'm clear - do we still have versions of these in the storybook?

The reasoning here was to remove them entirely, so they won't appear in the new Storybook going forward

joecowton1:
> > Just so I'm clear - do we still have versions of these in the storybook?
> 
> The reasoning here was to remove them entirely, so they won't appear in the new Storybook going forward

Ok think i misunderstood earlier - if the storybook is in many ways acting as the Source docs, do we not think it's important to surface these packages there?

ob6160:
> > > Just so I'm clear - do we still have versions of these in the storybook?
> > 
> > 
> > The reasoning here was to remove them entirely, so they won't appear in the new Storybook going forward
> 
> Ok think i misunderstood earlier - if the storybook is in many ways acting as the Source docs, do we not think it's important to surface these packages there?

The thinking behind it was that storybook is usually for building UI components and pages in isolation and as the two eslint packages are more developer facing it seemed a bit redundant to host their readme markdown files both in GitHub and as their own story.

I can see the argument that if the two libraries were visible in the Storybook discoverability would be improved. In https://github.com/guardian/csnx/pull/231 we also propose a general Source-wide readme in /docs which links to those two packages which might help with this?

joecowton1:
> > > > Just so I'm clear - do we still have versions of these in the storybook?
> > > 
> > > 
> > > The reasoning here was to remove them entirely, so they won't appear in the new Storybook going forward
> > 
> > 
> > Ok think i misunderstood earlier - if the storybook is in many ways acting as the Source docs, do we not think it's important to surface these packages there?
> 
> The thinking behind it was that storybook is usually for building UI components and pages in isolation and as the two eslint packages are more developer facing it seemed a bit redundant to host their readme markdown files both in GitHub and as their own story.
> 
> I can see the argument that if the two libraries were visible in the Storybook discoverability would be improved. In #231 we also propose a general Source-wide readme in /docs which links to those two packages which might help with this?

I'm leaning towards thinking we should include them, as much as that seems like a lot of work for a tiny thing. If its not blocking you maybe we can wait for @sndrs on this one?

ob6160:
> > > > > Just so I'm clear - do we still have versions of these in the storybook?
> > > > 
> > > > 
> > > > The reasoning here was to remove them entirely, so they won't appear in the new Storybook going forward
> > > 
> > > 
> > > Ok think i misunderstood earlier - if the storybook is in many ways acting as the Source docs, do we not think it's important to surface these packages there?
> > 
> > 
> > The thinking behind it was that storybook is usually for building UI components and pages in isolation and as the two eslint packages are more developer facing it seemed a bit redundant to host their readme markdown files both in GitHub and as their own story.
> > I can see the argument that if the two libraries were visible in the Storybook discoverability would be improved. In #231 we also propose a general Source-wide readme in /docs which links to those two packages which might help with this?
> 
> I'm leaning towards thinking we should include them, as much as that seems like a lot of work for a tiny thing. If its not blocking you maybe we can wait for @sndrs on this one?

yep, very much non blocking - happy to wait

PR #232, opened 2022-11-28
Dependency updates | 28th Nov 2022

ob6160:
## What are you changing?

- Bump the root dev dependencies 
- Leave `typescript` and `babel-loader` as they could affect consumers of libraries within csnx

## Why?

- We'd like to keep our dependencies up to date.


package.json
```diff
@@ -8,59 +8,59 @@
...
 		"sort-package-json": "2.1.0",
 		"ts-jest": "29.0.3",
 		"ts-node": "10.9.1",
 		"tsconfig-paths-webpack-plugin": "4.0.0",
-		"tslib": "2.4.0",
-		"type-fest": "3.1.0",
-		"typescript": "4.8.4",
+		"tslib": "2.4.1",
+		"type-fest": "3.3.0",
+		"typescript": "4.9.3",
```
joecowton1, line 63: I'd be cautious of this getting too far ahead of the libs / packages

package.json
```diff
@@ -8,59 +8,59 @@
...
-		"babel-loader": "8.2.5",
-		"eslint": "8.25.0",
-		"husky": "8.0.1",
-		"jest": "29.2.0",
-		"jest-environment-jsdom": "29.2.0",
-		"lint-staged": "13.0.3",
-		"nx": "15.0.13",
-		"prettier": "2.7.1",
+		"@typescript-eslint/parser": "5.44.0",
+		"babel-loader": "9.1.0",
```
joecowton1, line 46: do we know if this will work with other packages on older versions?

package.json
ob6160: I thought all of the packages (apart from `eslint-config-typescript` which is pinned to `typescript@4.2.2`) just used the workspace version. Confused about how this could be getting too far ahead, am I missing something?

ob6160: ah, really interesting problem for library maintainers, thanks for pointing it out! i'll put this back to where it was for now.

 it'd be good to have a wider discussion about how we decide when it's "allowed" to upgrade deps that could affect consumers of our libraries... it seems really unclear/ ill defined right now. Did some reading around it, an interesting issue: https://github.com/microsoft/TypeScript/issues/36207 and the project they mention that "downlevels" d.ts files.

package.json
ob6160: no we dont' - i'll revert this. i'd like to discuss a process for how we could work this out though as it seems really unclear to me. I spose the package upgrade policy usually is "major upgrades only when we need 'em" as i understand it

package.json
joecowton1: Maybe a good one to bring up at CSI meetings? I think its less a question of being allowed and more thinking in terms of balance between what upgrading deps gives us (security, bug fixes etc) vs the challenges it creates for the other platforms that consume these libs packages etc. 

If teams don't have time to upgrade TS on their platforms and as a result can't use Source we're shooting ourselves in the foot there so its always a compromise. Having everything all new and shiny means nothing if no one uses it :)

joecowton1:
Nice one for getting through these 👍

joecowton1 (approved):

PR #235, opened 2022-11-29
Add Source Storybook docs

tjsilver:
## What are you changing?

- We are adding the various stories that form the documentation for Source
- We have also added the stories for the source-eslint-plugin packages for now, though we might decide to have these as separate Storybooks in the future (in keeping with the 'Storybook-per-package' paradigm in csnx).
- Note for reviewers: We created a new app called `source-storybook` for this, but please suggest alternatives if you think this isn't the most logical/sensible place!

## Why?

- We need a place to capture these documents outside of the Source packages themselves.


apps/source-storybook/src/README.md
```diff
@@ -0,0 +1,3 @@
+# Docs
+
+Please see our [Storybook docs](https://guardian.github.io/source)
```
joecowton1, line 3: Is there a reason to link out of here rather than having the docs in this readme?

ob6160 (commented):
wondering if we need any of the following in the new `source-storybook` app dir as we have code in the `.storybook` dir that'll need type checking and eslint-ing (and maybe in src in the future?).. package.json might end up just having the project name in it, not sure if needed here:

- tsconfig
- package.json
- eslintrc

apps/source-storybook/src/components.stories.mdx
```diff
@@ -0,0 +1,9 @@
+import { Meta } from '@storybook/addon-docs';
+import ReadMe from '../../../libs/@guardian/source-react-components/README.md';
```
ob6160, line 2: I think I've found a way for us to import this using the package name.

- This line changed to: `import ReadMe from '@guardian/source-react-components/README.md';` 
- A package.json containing the following:
```json
{
	"name": "source-storybook",
	"devDependencies": {
		"@guardian/source-react-components": "workspace:*"
	}
}
```

works well for me locally 🙂

ob6160: hopefully the same for the other place we do this too

joecowton1: Yeah thats nicer

apps/source-storybook/src/README.md
tjsilver: Good point. That links to the Storybook that is currently on the old Source repo. We can look into what to replace it with when the time comes.

apps/source-storybook/src/components.stories.mdx
tjsilver: Great - you happy to commit those changes @ob6160 ?

ob6160: sure I’ll push it up in a bit

apps/source-storybook/src/README.md
tjsilver: I've deleted this file as it sits more in docs than in the Storybook app.

joecowton1 (approved):
This looks great 👌

PR #237, opened 2022-12-01
Flatten nested Source stories

ob6160:
## What are you changing?

- We previously nested stories for Source under `Packages/[package name]/*` when they were part of a combined Storybook in the original repository. This changes that scheme so we don't nest and the available packages are shown by default in the composed csnx Storybook.

## Why?

- This will make it easier for users to see which components are available to them at a glance. (thanks @joecowton1 for the suggestion)

Screenshot 2022-12-01 at 11 00 36


joecowton1:
Know this is only a draft, but could we go one step further and remove the package name from the title too? I think we still have one layer of nesting as it is.

joecowton1 (approved):
LGTM

PR #239, opened 2022-12-01
remove eslint-config-prettier from eslint-config

georgeblahblah:
## What are you changing?

- Removes `eslint-config-prettier` from `eslint-config`

## Why?

- I wanted to open a discussion around whether `eslint-config` should assume that the project will also be using `prettier`. `eslint-config-prettier` turns off some core `eslint` rules which some projects using `eslint-config` might want to keep.

### DIFF
--- a/libs/@guardian/eslint-config/index.js
+++ b/libs/@guardian/eslint-config/index.js
@@ -8,7 +8,6 @@ module.exports = {
 		'eslint:recommended',
 		'plugin:import/errors',
 		'plugin:import/warnings',
-		'prettier',
 	],
 	plugins: ['eslint-comments'],
 	rules: {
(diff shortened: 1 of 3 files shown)


sndrs:
I'm curious to know why you would not want to use prettier?

I think there's a lot of value in standardising things like formatting. For example, one change that consumers of our `@guardian/prettier` config got for free was this change https://github.com/guardian/configs/pull/220.

Prettier formats a lot more than just JS/TS. So by using it we get consistent, well-formatted (in the example above, more accessibly-formatted) files for JavaScript, JSX, Angular, Vue, Flow, TypeScript, CSS, Less, SCSS, HTML, Ember/Handlebars, JSON, GraphQL, Markdown (including GFM and MDX) and YAML.

There are many [3rd party plugins](https://prettier.io/docs/en/plugins.html#community-plugins) available too, for things like astro, elm, kotlin, svelte, bash etc.

fyi, this is their rationale https://prettier.io/docs/en/comparison.html for not formatting in linting.

mxdvl:
Yes, after some further thought I agree we should keep Prettier

mxdvl:
Whoops I didn’t open this PR

tjsilver:
Closing again as it looks like it was re-opened in error.

PR #252, opened 2022-12-06
Migrate final versions of Source files

tjsilver:
## What are you changing?

- Some files have changed since we first copied them in from the Source repo. This PR brings over any changes so that the repos are up-to-date.

## Why?

- We can then archive the original Source folder and publish Source packages to NPM from the monorepo.


sndrs (approved):
great work!!!

PR #286, opened 2022-12-06
add age warning component to kitchen

marjisound:
## What are you changing?
- This PR is adding the `AgeWarning` component which inserts a yellow warning banner on the page showing how old the content is. 
image

supporting dark mode:
image

Note: Although for the time being there's no  dark mode support in `csnx`, the `darkModeCss` function was added in the styles.ts for AgeWarning component to make it usable for the apps. Whenever we come up with a solid solution for dark mode, this component can also be updated accordingly. 

## Why?
- This component is used in DCR already. But because there's also a usecase for it in AR, we can probably benefit from adding it to kitchen

### DIFF
--- a/libs/@guardian/source-react-components-development-kitchen/src/age-warning/AgeWarning.stories.tsx
+++ b/libs/@guardian/source-react-components-development-kitchen/src/age-warning/AgeWarning.stories.tsx
@@ -31,3 +31,6 @@ emptyWarningPrefix.args = { age: '5 years', warningPrefix: '' };
 
 export const customWarningPrefix = Template.bind({});
 customWarningPrefix.args = { age: '5 years', warningPrefix: 'This book is ' };
+
+export const supportsDarkMode = Template.bind({});
+supportsDarkMode.args = { age: '10 years old', supportsDarkMode: true };
--- a/libs/@guardian/source-react-components-development-kitchen/src/age-warning/AgeWarning.stories.tsx
+++ b/libs/@guardian/source-react-components-development-kitchen/src/age-warning/AgeWarning.stories.tsx
@@ -1,0 +1,33 @@
+import type { Story } from '@storybook/react';
+import type { AgeWarningProps } from './AgeWarning';
+import { AgeWarning } from './AgeWarning';
+
+export default {
+	component: AgeWarning,
+	title: 'AgeWarning',
+};
+
+const Template: Story<AgeWarningProps> = (args: AgeWarningProps) => (
+	<AgeWarning {...args} />
+);
+
+export const ageWarning = Template.bind({});
+ageWarning.args = { age: '10 years old' };
+
+export const smallWarning = Template.bind({});
+smallWarning.args = { age: '5 months old', size: 'small' };
+
+export const screenReaderVersion = Template.bind({});
+screenReaderVersion.args = {
+	age: '20 million years old',
+	isScreenReader: true,
+};
+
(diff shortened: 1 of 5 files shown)


libs/@guardian/source-react-components-development-kitchen/src/age-warning/AgeWarning.stories.tsx
```diff
@@ -0,0 +1,36 @@
+import type { Story } from '@storybook/react';
```
abeddow91, line 1: nice story coverage!

abeddow91 (approved):
Nice one! 🚢

sndrs (approved):

sophie-macmillan (approved):

libs/@guardian/source-react-components-development-kitchen/src/age-warning/AgeWarning.tsx
```diff
@@ -0,0 +1,97 @@
...
+	visuallyHidden,
+} from '@guardian/source-foundations';
+import { darkModeCss } from './styles';
+
+export interface AgeWarningProps {
+	/**
+	 * the age to be shown in the warning. e.g. 5 years
+	 */
+	age: string;
+	/**
```
tkgnm, line 17: Looks amazing Marjan. Do we want to add some sort of logic so that if passed a unix time stamp the component could format the age appropriately?

marjisound: Thanks Tom for looking, the age prop is a string that would be shown in the warning as it is. For example if the age is `5 years` and the warningPrefix is not given, the message that's shown is `This article is more than 5 years old`

libs/@guardian/source-react-components-development-kitchen/src/age-warning/AgeWarning.tsx
```diff
@@ -0,0 +1,97 @@
...
+	 * size of the yellow warning banner
+	 */
+	size?: 'small' | 'medium';
+	/**
+	 * the message to be shown before the age. default is 'This article is more than '
+	 */
+	warningPrefix?: string;
+	/**
+	 * use this if platform support dark mode
```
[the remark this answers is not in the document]
marjisound: Done

abeddow91 (approved):

sndrs (approved):

libs/@guardian/source-react-components-development-kitchen/src/age-warning/AgeWarning.tsx
tkgnm: Gotcha. Sorry - what I meant was that instead of passing in a string for the age, you could pass in a date and the component would figure out what the string should be. Essentially, we move the logic you have wrote in your AR pull request to this one: https://github.com/guardian/dotcom-rendering/pull/6686/files#diff-6c405a7ab45efe04c2c48734531cdcd4a51dae5ba56f381ab75afddd0bc79297R14

That could then mean that both DCR and AR will use exactly the same logic for figuring out dates, which could reduce code and enable consistency? Just a thought.

marjisound: Great suggestion @tkgnm 
For visibility, we had a group chat about this, and decided to create a health ticket for this improvement. So hopefully soon we'll move the age calculation logic to the AgeWarning component :)

PR #295, opened 2022-12-07
move some package from root to projects

sndrs:
## What are you changing?

- move some dev deps from the root into the packages that need them

## Why?

- packages should declare what they need, the root should be for root-level stuff only

n.b. this excludes storybook currently, because it breaks. we should probably have our own storybook runner (as with linting and bundling) that installs what it needs but that's beyond the scope of this


package.json
```diff
@@ -68,13 +58,8 @@
...
 		"peerDependencyRules": {
 			"ignoreMissing": [
-				"@babel/core",
-				"@emotion/react",
-				"eslint",
-				"prettier",
-				"react",
-				"react-dom",
-				"typescript"
+				"@types/node",
```
tjsilver, line 61: How do we decide what to ignore here?

sndrs: these are things are inherently repo-wide ( think). e.g. node version can only be the version in the root nvmrc, so this `@types/node` will always be this version.

joecowton1 (dismissed):

joecowton1 (approved):

PR #296, opened 2022-12-07
Specify minimum TS support in packages

sndrs:
## What are you changing?

- adds `typescript` to NPM package `peerDependencies` (and so `devDependencies`)
- makes `peerDependencies` of typescript optional using [`peerDependenciesMeta`](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta)

### Example

```json
"peerDependencies": {
	"typescript": "^4.3.2"
},
"peerDependenciesMeta": {
	"typescript": {
		"optional": true
	}
}
```

## Why?

- for packages that provide types, we need to be able to declare the minimum version of typescript that they work with
- but just because you provide types doesn't mean the code is being used in typescript, so typescript can't be a standard peer dep
- using `peerDependenciesMeta` means that if typescript is missing, package managers shouldn't complain, but if they find an incompatible version, they will


joecowton1 (approved):
Great shout

sndrs:
> How was 4.3 chosen? Which specific feature was only introduced then?

it was the lowest version that built ok :)

PR #302, opened 2022-12-08
Remove redundant source storybook config

tjsilver:
## What are you changing?

- Taking out some unused parameters from the Source Storybook configs.

## Why?

- We had to duplicate the `.storybook` folders from a single one to three for the migration to CSNX. Some of the config is unused in some packages.


ob6160 (approved):
nice work, way cleaner 🧹

PR #311, opened 2022-12-12
Sophie m/add file input

sophie-macmillan:
## What are you changing?
Adds a new component to source-react-components-kitchen, `FileInput`.
This is an input of type file which is styled as a button. If the `optional` prop is passed, on selecting a file a remove file button is rendered, if it is not optional, once selected, the file cannot be removed, only changed.

## Why?
The new callout component requires a file input field, so by extracting it to source kitchen we can insure it is consistent between the 2 repos.

## Screenshots
**Pre Upload**
image

**Optional**
image

**Required**
image

**With Custom Error**
image

**With FileType Error**
image

**With FileSize Error**
image


libs/@guardian/source-react-components-development-kitchen/src/file-input/types.ts
```diff
@@ -0,0 +1,36 @@
+export interface FileInputProps {
+	id?: string;
```
tkgnm, line 2: This is so nice and clear Sophie - thanks for making it so easy to read! Do we want to be able to pass in `format` so that the button can be styled accordingly? News, opinion etc.

sophie-macmillan: I think we should just be able to override the primary colour in the theme to do that! Seems to be the pattern in source-react-components anyway...

tkgnm: Okay cool - if at all possible, could we jump on a call just so you can show me how I'd do that for my callout PR?

libs/@guardian/source-react-components-development-kitchen/src/file-input/FileInput.stories.tsx
```diff
@@ -0,0 +1,50 @@
+import type { ReactElement } from 'react';
+import { FileInput } from './FileInput';
+
+/**
+ * [Storybook](https://guardian.github.io/csnx/?path=/story/source-react-components-development-kitchen_fileinput--optional-file-input&globals=backgrounds.value:transparent
+ * [GitHub](https://github.com/guardian/csnx/tree/main/libs/@guardian/source-react-components-development-kitchen/src/file-input
+ * [NPM](https://www.npmjs.com/package/@guardian/source-react-components)
```
tjsilver, line 7: I think this should be `.../source-react-components-development-kitchen`

sophie-macmillan: Good spot - have updated!

libs/@guardian/source-react-components-development-kitchen/src/file-input/FileInput.stories.tsx
```diff
@@ -0,0 +1,50 @@
+import type { ReactElement } from 'react';
+import { FileInput } from './FileInput';
+
+/**
+ * [Storybook](https://guardian.github.io/csnx/?path=/story/source-react-components-development-kitchen_fileinput--optional-file-input&globals=backgrounds.value:transparent
+ * [GitHub](https://github.com/guardian/csnx/tree/main/libs/@guardian/source-react-components-development-kitchen/src/file-input
```
tjsilver, line 6: Lines 5 and 6 lines require a closing ')' for the links to work properly.

libs/@guardian/source-react-components-development-kitchen/src/file-input/stringifyFileBase64.ts
```diff
@@ -0,0 +1,30 @@
...
+			() => {
+				// remove data:*/*;base64, from the start of the base64 string
+				const fileAsBase64 = reader.result?.toString().split(';base64,')[1];
+
+				if (fileAsBase64) {
+					resolve(fileAsBase64);
+				} else {
+					reject(
+						new Error(
+							'Sorry there was a problem with the file you uploaded above. Check the size and type. We only accept images, pdfs and .doc or .docx files',
```
tjsilver, line 15: Does the bit about .doc and .docx need to be removed from lines 15 and 25, or added to the list of accepted files?

libs/@guardian/source-react-components-development-kitchen/src/file-input/FileInput.tsx
```diff
@@ -0,0 +1,90 @@
...
+	hideLabel,
+	supporting,
+	error,
+	onError,
+	onUpload,
+	...props
+}) => {
+	const [chosenFile, setChosenFile] = useState<null | string>();
+
+	const getFileName = (filepath?: string): string =>
```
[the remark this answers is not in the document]
sophie-macmillan: This is a good suggestion. I've done that now!

libs/@guardian/source-react-components-development-kitchen/src/file-input/stringifyFileBase64.ts
sophie-macmillan: I've updated the way this works. The component now takes a `validFileType` which defaults to `['image/png', 'image/jpeg', 'image/jpg']`, and an optional `maxFileSize`, which are both used for additional information, and to construct the appropriate error message when an incorrect file is uploaded.

libs/@guardian/source-react-components-development-kitchen/src/file-input/FileInput.tsx
```diff
@@ -0,0 +1,112 @@
...
+				<>
+					{optional && (
+						<button
+							css={(theme: Theme) => customUpload(theme.fileInput)}
+							onClick={onRemoveFile}
+						>
+							Remove File
+						</button>
+					)}
+					<span>{getFileName(fileName)}</span>
```
tjsilver, line 107: can this just be `filename` now that it's in state?

sophie-macmillan: It can - sorry missed this!

libs/@guardian/source-react-components-development-kitchen/src/file-input/FileInput.tsx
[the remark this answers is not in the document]
tjsilver: I was thinking that this function would be made redundant by retrieving the filename directly from the API?

tjsilver (approved):
Tested locally and all works well - lots of story coverage too - great work! 💯

PR #313, opened 2022-12-14
Add isNonNullable type guard

mxdvl:
## What are you changing?

- Add a type guard to check whether a value is neither `null` nor `undefined`: `isNonNullable`

## Why?

- This is a very helpful tool to narrow types with filtering:

```ts
/** @type {string[]}, as only non-nullable values are kept */
const strings = ["hello", undefined, "world", null].filter(isNonNullable)
```


sndrs (changes requested):
this will [only work with `typescript@>=2.8`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html#predefined-conditional-types), so we'd need to bump the peerdep too. 

how would this affect current consumers?

mxdvl:
> this will [only work with `typescript@>=2.8`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html#predefined-conditional-types), so we'd need to bump the peerdep too.
> 
> how would this affect current consumers?

I think [we already require `^4.3.2`](https://github.com/guardian/csnx/blob/24c5a9b53cc1b3e3cc2f9a808649ede31d85ba85/libs/@guardian/libs/package.json#L17).

sndrs:
oh god i was not with it, i took 2.8 to mean 4.8 🤦

PR #316, opened 2022-12-14
Fix lint error on any type

sophie-macmillan:
## What are you changing?
Changes any any type for the param passed through to the onError prop in the file input.


sndrs (approved):
love it - sorry it wasn't caught, looking into that now

tjsilver (approved):

PR #317, opened 2022-12-14
Enforce `no-explicit-any` in `@guardian/eslint-config-typescript`

sndrs:
## What are you changing?

- disallow explicit use of `any`

## Why?

- `any` is basically no type-checking at all, and we if we're going to use types, we should make sure we do (currently it's a warning)


tjsilver (approved):
Good idea! 👍

PR #327, opened 2022-12-19
Add atoms-rendering Amplify instance to composed storybook

joecowton1:
## What are you changing?

Adds Atoms-Rendering storybook to composed storybook.


tjsilver (changes requested):
I think the title needs to be updated

joecowton1:
> I think the title needs to be updated

fixed now, thanks

tjsilver (approved):
👍

PR #328, opened 2022-12-19
Tweaks to styling after UX review

sophie-macmillan:
## What are you changing?
Updates the border thickness, and the filePath font for the FileInput component.

**Before**
image

**After**
image

## Why?
This component has been reviewed by UX, so we are making the needed style changes
-


tjsilver (approved):
Code looks good - would be great to see some stories or screenshots of the small/xmall sizes. I tested them locally and all looks fine.

libs/@guardian/source-react-components-development-kitchen/src/file-input/FileInput.stories.tsx
```diff
@@ -64,6 +64,17 @@ const fileSizeValidation = (): ReactElement => {
 	);
 };
 
+const smallFileInput = (): ReactElement => {
```
tjsilver, line 67: There seem to be three button sizes, IIRC: `default`, `small` and `xmall`. It is slightly confusing that this button is called `smallFileInput` but actually refers to an `xsmall` button. It would be good to have one of each, or make it a bit clearer?

tjsilver (approved):
Great work!

PR #332, opened 2022-12-23
Update browserslist stats

SiAdcock:
## What are you changing?

- Update browserslist to v4.21.6
- Update browserslist-stats.json and README with latest data from GA
- Add some opinions to the readme regarding the datatset

## Why?

@arelra and I were hoping to filter out some legacy browsers – notably Safari 10, which doesn't support dynamic imports. Sadly this hasn't happened 😭

Upgrading browserslist has picked up explicit support for browsers launched in the last 6 months. It's unclear if this will make any difference to bundles, but it might be useful from the perspective of someone checking the README for a canonical list of browsers we support.

It also drops support for older versions of Android-based browsers and Opera.

The "Report content" section adds some suggestions for how to filter the dataset in Google Analytics. We feel it's important that the report is generated using consistent dataset options. The values here seem like sensible choices, but they are not strongly-held opinions, so please do offer alternative suggestions.


joecowton1 (approved):
Thanks Simon! Looks great. Does this need a changeset do you think?

SiAdcock:
>  Does this need a changeset do you think?

Oh yes, good shout @joecowton1! I've major-bumped it as the `browserslist` peer dependency is changing. Hope this makes sense 😊

arelra:
Just an fyi when this is bumped in DCR this unit test on DCR will fail:

https://github.com/guardian/dotcom-rendering/blob/1605a1919fbb6ab0f4d79375649fbccc7904489c/dotcom-rendering/scripts/webpack/browser-targets.test.ts#L7

I did this deliberately as a heads up to the browser targets changing. I'll add a comment to the test in DCR.

joecowton1 (approved):

joecowton1:
Thanks both, good to know.

PR #340, opened 2023-01-04
Add readme and make script for nginx server

joecowton1:
## What are you changing?

- copying in readme and changelog from original atoms-rendering
- adding a make script for the nginx server setup


libs/@guardian/atoms-rendering/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
ob6160, line 1: the license at the root of the monorepo (https://github.com/guardian/csnx/blob/main/LICENSE) looks like it's the same as this one. Does the root one cover this subdirectory too I wonder? Or would it just be better to be explicit and include it twice 🤔

ob6160 (dismissed):
lgtm :shipit:

libs/@guardian/atoms-rendering/LICENSE
joecowton1: Good spot, removed.

PR #342, opened 2023-01-05
Refactor countries to use inferred types

mxdvl:
## What are you changing?

- rely on the `as const` assertion of countries to infer CountryCode and CountryKey types

## Why?

- less duplication
- clearer relation between the definition and the type


sndrs:
i guess by having the `CountryCode` type, it ensures that `countries#countryCode` is a valid string.

whether that's going to catch an error is a good question though (`CountryCode` just as likely to contain an invalid string)

.changeset/friendly-apes-fry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/libs': patch
+---
+
+Refactor types to be inferred from TS as const assertion
```
sndrs, line 5:
```suggestion
Refactor `CountryCode` type to infer the possible values from `countries#countryCode`.
```

mxdvl:
I think that the closest colocation of code, short name and long name is the one that’s least prone to introducing errors.

If we want to add more security, the only thing I could think of is downloading a dataset from an authoritative source.
- Wikipedia? https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2
- ISO (from archive.org)? https://web.archive.org/web/20160804224459/http://www.iso.org/iso/home/standards/country_codes/country_names_and_code_elements_xml-temp.htm
- CIA? https://www.cia.gov/the-world-factbook/references/country-data-codes/

sndrs (dismissed):

joecowton1 (approved):

PR #344, opened 2023-01-06
loadScript rejects with more helpful error messages

### LINKED ISSUE #333: loadScript rejects with an event, but it should reject with an error
**Describe the bug**
When `loadScript` fails, it rejects with an `Event` object, rather than an `Error` as stated in the [readme](https://github.com/guardian/csnx/blob/main/libs/%40guardian/libs/src/loadScript/README.md#loadscriptsrc-props). 

The consequence of this is a large number of errors that get reported in Sentry as `[object Event]
_promiseRejectionHandler([native code])`<sup>1</sup> ([example](https://sentry.io/organizations/the-guardian/issues/2778776754/events/e31724a4449b4a288bc1e0837bd31744/?project=35463&query=is%3Aunresolved&referrer=previous-event)).

I believe this is happening because the [script's `onerror` event handler is set to the `reject` callback](https://github.com/guardian/csnx/blob/b09f398661b8a552d68d26a4074c60cbf14e35a1/libs/%40guardian/libs/src/loadScript/loadScript.ts#L31) in the Promise. When the [`onerror`](https://developer.mozilla.org/en-US/docs/Web/API/Window/error_event) handler is fired, it passes its arguments to the `reject` callback. This handler receives the Event that caused the error as the first argument. This is what is passed to `reject`.

**To Reproduce**
Steps to reproduce the behaviour:

This is a work in progress 😅

**Expected behaviour**

If there is an error loading a script, `loadScript` should reject with an `Error` object with a meaningful error message (wherever possible) rather than an `Event`.

**Footnotes**

<sup>1</sup> Often when these errors are reported, the message `GA - error: [object Event]` is printed to the console. This console log originates in DCR, when it [uses `loadScript` to load Google Analytics](https://github.com/guardian/dotcom-rendering/blob/6e3004efa9bc2477d2351da9dca1b3332cb78daa/dotcom-rendering/src/web/browser/bootCmp/init.ts#L135).

SiAdcock:
## What are you changing?

Make changes to the way the `loadScript` promise gets rejected. The objective is to provide more helpful and more granular error messages. There are [a large number of issues in Sentry](https://sentry.io/organizations/the-guardian/issues/?project=35463&query=is%3Aunresolved+%5Bobject+Event%5D&referrer=issue-list&sort=freq&statsPeriod=14d) with the message `[object Event]`, which this change attempts to address.

The prioritised ways in which the promise may be rejected is as follows:

1. if the script's [`onerror`](https://developer.mozilla.org/en-US/docs/Web/API/Window/error_event) handler receives an `Error` object, we reject with this
1. if the handler receives an event `string`, we assume the string is a human-readable error message describing the error, and reject with this
1. if the handler receives an `Event` object, we reject with the `src` attribute from the target element (which we assume is the script element)
1. otherwise, we reject with a generic error message

### Questions for reviewers to consider

1. Is the priority order of the conditions correct?
1. Are there additional ways of getting a useful error message that are not covered here?
1. Are any of the above ways of getting an error message invalid or unhelpful?
1. In 3 of the 4 conditions described above, we reject with a string, and in one case we reject with an Error object. Should we always be rejecting with an Error object?

## Why?

Fixes #333


joecowton1 (approved):
Thanks Simon! I don't have much to add but this looks great.

PR #356, opened 2023-01-10
Add Atoms-Rendering types output dir to tsconfig

joecowton1:
## What are you changing?

- adding explicit types exports from atoms-rendering

## Why?

- DCR is currently using types accessed from the dist folder is an odd way. By explicitly exporting these types we will be able to import them properly instead.


sndrs (approved):
i paired on this and approve of it

ob6160 (approved):

PR #360, opened 2023-01-12
Add data-ignore attribute to KnowledgeQuiz atoms <ol> to avoid duplicate list numbers 

Georges-GNM:
## What are you changing?

Knowledge quizzes are currently showing a duplicated number before the question:

image

This css change removes it in when the atom gets rendered by DCR (the change should only affect atoms in the context of DCR; if there are any other consumers of the package then they shouldn't see any difference in behaviour from this update.

image

Paired with @bryophyta and @ioannakok on this - the issue is caused by css defined in DCR, which raised the question of whether this issue could be solved through changes in that repo, e.g. telling DCR to not apply its global css stylings to atoms imported from atoms-rendering. However, this requires a larger conversation with the Web Experience team; hence this simple change to solve the problem now, until we can have that discussion.

### DIFF
--- a/libs/@guardian/atoms-rendering/src/KnowledgeQuiz.tsx
+++ b/libs/@guardian/atoms-rendering/src/KnowledgeQuiz.tsx
@@ -82,7 +82,7 @@ export const KnowledgeQuizAtom = ({
 					/>
 				</div>
 			)}
-			<ol>
+			<ol data-ignore="global-ol-styling">
 				{questions.map((question, idx) => (
 					<Question
 						key={question.id}
@@ -149,9 +149,6 @@ export const Question = ({
 		<li
 			css={css`
 				${theme === ArticleSpecial.Labs ? textSans.medium() : body.medium()};
-				::before {
-					content: none !important;
-				}
 			`}
 		>
 			<fieldset css={fieldsetStyle}>
--- a/libs/@guardian/atoms-rendering/src/KnowledgeQuiz.tsx
+++ b/libs/@guardian/atoms-rendering/src/KnowledgeQuiz.tsx
@@ -149,6 +149,9 @@ export const Question = ({
 		<li
 			css={css`
 				${theme === ArticleSpecial.Labs ? textSans.medium() : body.medium()};
+				::before {
+					content: none !important;
+				}
 			`}
 		>
 			<fieldset css={fieldsetStyle}>
--- a/.changeset/shaggy-ways-love.md
+++ b/.changeset/shaggy-ways-love.md
@@ -1,0 +1,5 @@
+---
+'@guardian/atoms-rendering': patch
+---
+
+Adding data-ignore attribute to KnowledgeQuiz atom ordered lists to remove duplicate list numbers
--- a/.changeset/spotty-wasps-leave.md
+++ b/.changeset/spotty-wasps-leave.md
@@ -1,0 +1,5 @@
+---
+'@guardian/atoms-rendering': minor
+---
+
+Added css override to KnowledgeQuiz atom to avoid duplicate list numbers
--- a/.changeset/spotty-wasps-leave.md
+++ b/.changeset/spotty-wasps-leave.md
@@ -1,5 +1,0 @@
----
-'@guardian/atoms-rendering': minor
----
-
-Added css override to KnowledgeQuiz atom to avoid duplicate list numbers


ob6160 (approved):

libs/@guardian/atoms-rendering/src/KnowledgeQuiz.tsx
```diff
@@ -82,7 +82,7 @@ export const KnowledgeQuizAtom = ({
 					/>
 				</div>
 			)}
-			<ol>
+			<ol data-ignore="global-ol-styling">
```
OllysCoding, line 85: I wonder if this is the correct approach to take, as I understand:

DCR Removes the default `<ol>` styles with the CSS reset, then adds its own `<ol>` styles back manually.

---

However, that leaves DCR in a neutral position; atoms-rendering is actually adding the number twice, the first with the use of the <ol> which browsers style by default with numbers. Then, we add our own numbers to the DOM here: https://github.com/guardian/csnx/blob/main/libs/%40guardian/atoms-rendering/src/KnowledgeQuiz.tsx#L165

I think we should consider either:
1. Not adding our own numbers to the DOM on atoms-rendering, and trusting the platforms & browser to add these.
2. We may need to keep this data-ignore, but might want to consider also writing our own CSS to remove the list-styles, so that we're not relying on DCR to do so. 

Happy to chat a bit more about it! Perhaps the @guardian/csnx team may also have some insight on this, as it is a cost we've paid in having global styles in DCR

bryophyta: I thought it was the case that atoms-rendering had a reset rule for `<ol>` to remove the numbering, but that DCR's global style had a higher specificity and so overrode a-r's reset?

mxdvl: For the record, the relevant CSS selector for this is [`::marker`](https://developer.mozilla.org/en-US/docs/Web/CSS/::marker)

bryophyta: Sorry I realise that the above comment wasn't super clear. My thought is that removing the numbers from the Quiz atom would be a breaking change for atoms-rendering, because the numbers are there in the storybook for this component for instance. (Although arguably our PR that introduced the `<ol>` element in the first place should have been considered a breaking change; I think we weren't thinking that way because we were assuming the presence of resets.)

To my mind the 'bug' is in DCR's use of atoms-rendering, so I agree that this isn't an _ideal_ fix, but seemed like the best quick fix given the aim of fixing DCR without a breaking change to A-R. But this discussion has helped me to realise that my assumption about the contract with consumers of this package is that they will only override A-R's resets intentionally, and that might not be a reasonable assumption! Would be interested to hear your thoughts @OllysCoding  and anyone else!

mxdvl: I think the best fix here would be to add an emotion style to the list item which enforces a specific style, rather than relying on global resets. So either:
- applying `list-style-type: none;` to the `ol`
- applying `::marker { content: none; }` to the `li`
- removing the number from JSX and letting the browser do its thing… with potential slight variations

bryophyta: I don't mind any of those three options in principle, but just to (hopefully) clarify my point from above: these won't solve the current problem without a corresponding change in DCR because DCR's global CSS (not the reset, but [the `globalOlStyles`](https://github.com/guardian/dotcom-rendering/blob/e2d3b32346d15e89817f842538d96e1d81a019fa/dotcom-rendering/src/web/components/ArticleBody.tsx#L58)) are liable to override these three solutions in A-R

bryophyta: ps. I'm glad to see that `::marker` has decent support now! Last time I looked (a while ago) I think it was pretty low still!

https://caniuse.com/?search=marker

Georges-GNM: > consumers of this package is that they will only override A-R's resets intentionally

That still seems to be what's happening here? The "intention" in DCR is that all `<ol>` it renders have a particular styling, which would naturally applies to elements which may be imported in DCR. 

As @bryophyta  pointed out, the issue is still present even when trying to implement those suggestions in A-R. My understanding is that the number we really want to remove is getting generated from DCR, but implementing a solution there feels like it requires a larger discussion about how DCR interacts with A-R; in the meantime I am inclined to get this working fix out.

mxdvl: The problem that this introduces is a loose coupling between atoms-rendering and DCR. If you add styles to the `ol` in atoms-rendering, instead of a class name, these styles will have higher specificity than DCR and will therefore be applied.

I would suggest you apply `list-style-type: none;` rather than the class name to the `ol` in the repo.

bryophyta: @mxdvl I think we might be slightly talking past each other on the coupling issue because we have (I _think_) different understandings of what the current CSS situation is on the ground between A-R and DCR. Shall I put in some time for the people on this issue (optionally!) to chat it through synchronously?

mxdvl: >     * applying `list-style-type: none;` to the `ol`
>     * applying `::marker { content: none; }` to the `li`

Ah, yes, I did the mistake of assuming that `dotcom-rendering` was doing things in a standard way by setting `list-style-type`, but it is not and in fact relies on the `::before` pseudo-element. My first two suggestion should have been:

- applying `ol li::before { display: none; }` to the `ol`
- applying `::before { display: none; }` to the `li`

But I still wonder why we’re departing from browsers standards and would recommend investigating that in DCR in the first place!

PR #364, opened 2023-01-16
Tweak expanding wrapper styling, export the default themes and allow cssOverrides

sophie-macmillan:
## What are you changing?
This change updates the Expanding Wrapper

- Updates the button styling
- Allows cssOverrides to be passed
- Exports the default themes for the component

**Before**
image

**After**
image


joecowton1 (approved):
Nice one!

PR #365, opened 2023-01-16
Feat: extend article design types to include Profile and Timeline

abeddow91:
## What are you changing?
This PR extends the `ArticleDesign` type to include `Profile` and `Timeline`

## Why?
This change is due to an editorial desire to separate profile and timeline articles from the standard design, allow for specific styling of these articles

### DIFF
--- a/.changeset/violet-shirts-serve.md
+++ b/.changeset/violet-shirts-serve.md
@@ -1,0 +1,5 @@
+---
+'@guardian/libs': minor
+---
+
+Extends format's articleDesign type to include `Profile` and `Timeline`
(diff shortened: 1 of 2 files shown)


joecowton1 (approved):

joecowton1:
Thanks @abeddow91 ! Just to to double check this isn't a breaking change is it?

abeddow91:
> Thanks @abeddow91 ! Just to to double check this isn't a breaking change is it?

Hey @joecowton1 - I don't think so, this is only extending the type, not removing anything so its backwards compatible. My thought was then that it's just a minor but let me know if you think otherwise and I'll amend

joecowton1:
> > Thanks @abeddow91 ! Just to to double check this isn't a breaking change is it?
> 
> Hey @joecowton1 - I don't think so, this is only extending the type, not removing anything so its backwards compatible. My thought was then that it's just a minor but let me know if you think otherwise and I'll amend

Sounds good to me if thats the case :)

PR #371, opened 2023-01-18
Add `@guardian/ab-react` library

mxdvl:
## What are you changing?

- Move @guardian/ab-react from https://github.com/guardian/ab-testing/tree/main/packages/ab-react
- Updated the tsconfig to prevent the `@emotion/react` from being selected as a `jsxImportSource`.

## Why?

- Publishing packages from the monorepo is easier – or even possible at all
- The original package has had a chronic lack of improvements
- We’ve already moved `@guardian/ab-core` in #341


joecowton1 (dismissed):

mxdvl:
Sorry – the output of the build step was broken, and I fixed it in the latest change.

mxdvl:
I’ve tried to investigate the build error… and it has been very confusing to reproduce. It seems to only happen if the `@guardian/source-react-components-development-kitchen` are already built and present in the `dist` folder, which seem to suggest that there’s some conflicting types declarations somehow.

joecowton1:
> I’ve tried to investigate the build error… and it has been very confusing to reproduce. It seems to only happen if the `@guardian/source-react-components-development-kitchen` are already built and present in the `dist` folder, which seem to suggest that there’s some conflicting types declarations somehow.
> 
> _Exempli gratia_: [it took two runs for the `make build` step to pass](https://github.com/guardian/csnx/actions/runs/3953812358/jobs/6770515414). The difference between the two runs is that I locally ran the following commands `make clean && pnpm nx run @guardian/ab-react:build && make build` to populate the remove cache with a working version.

With this in mind is it safe to proceed or do you want to dig further into that one first?

joecowton1 (approved):
Nicely done 👍

PR #372, opened 2023-01-18
Update versions

joecowton1:
## What are you changing?

-

## Why?

-


ob6160 (approved):

mxdvl:
Would this not result in a breaking change for consumers? Technically, changing the peer dependencies might mean a consumer is no longer able to satisfy it.

joecowton1:
> I think it’s @sndrs who made me realise this should probably be a breaking change/major bump. I think changing the peer dependencies might mean a consumer is no longer able to satisfy it.

Could you expand Max - do you have an example of where this would break?

mxdvl:
> > I think it’s @sndrs who made me realise this should probably be a breaking change/major bump. I think changing the peer dependencies might mean a consumer is no longer able to satisfy it.
> 
> Could you expand Max - do you have an example of where this would break?

Say in `app` you have `@guardian/libs@^10` and `@guardian/core-web-vitals@2.0.2`, you cannot simply update to `@guardian/core-web-vitals@2.0.3` without also bumping `@guardian/libs@^12`.

In practice, we’re doing this because we know that 100% of our consumers actually want to have the latest peer dependency, so it might be fine, but we might want to prevent this from happening in the future, by creating a script that prevents anything but a major bump when changing the peer dependencies of a package.

PR #373, opened 2023-01-18
Enforce sequencing of builds

mxdvl:
## What are you changing?

- add a dependency to `@guardian/ab-react` in `@guardian/source-react-components-development-kitchen`

## Why?

- `@guardian/ab-react` has to be built before `@guardian/source-react-components-development-kitchen`, as the existence of the latter’s build output in the `dist/` folder generates the following error when building the former.
- https://github.com/guardian/csnx/pull/371#issuecomment-1396223148


mxdvl:
This seems like a very bad way to do coupling.

PR #374, opened 2023-01-19
Fix `@guardian/ab-react` builds

mxdvl:
## What are you changing?

- Make the `@guardian/ab-react` project load Emotion, so there is no JSX type declarations between React and Emotion 

## Why?

- Alternative to #373


mxdvl:
Cherry-picked 02f8ac11c7748aeaf5992575f5e8f3b5ea41c90e

PR #375, opened 2023-01-19
Update background of GuideAtoms on Analysis articles

sophie-macmillan:
## What does this change?
Updates the GuideAtom to optionally take a design, this is because Analysis articles have a light pink background, so we need to use a different Guide background for these pieces.

## Why?
The grey on pink background didn't look right so we had designs to update it.

## Screenshots

| Before      | After      |
|-------------|------------|
| (in article) image | (in storybook) image |

### DIFF
--- a/.changeset/chilled-icons-taste.md
+++ b/.changeset/chilled-icons-taste.md
@@ -1,0 +1,5 @@
+---
+'@guardian/atoms-rendering': patch
+---
+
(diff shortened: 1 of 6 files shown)


joecowton1 (dismissed):
Nicely done 👍

joecowton1 (approved):

PR #390, opened 2023-01-31
Update tslib version

joecowton1:
## What are you changing?

- updating tslib versions across csnx

## Why?

- we were seeing per deps errors

### DIFF
--- a/.changeset/loud-games-suffer.md
+++ b/.changeset/loud-games-suffer.md
@@ -1,7 +1,0 @@
----
-'@guardian/eslint-plugin-source-foundations': patch
-'@guardian/eslint-plugin-source-react-components': patch
-'@guardian/source-react-components-development-kitchen': patch
----
-
-Bump tslib to v2.4.1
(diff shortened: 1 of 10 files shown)


mxdvl (commented):
I'm not sure why a devDependencies update should result in a patch update. I believe it should not release a new version at all.

joecowton1:
> I'm not sure why a devDependencies update should result in a patch update. I believe it should not release a new version at all.

Can you suggest another way to consume the updated packages than via a release?

mxdvl:
> > I'm not sure why a `devDependencies` update should result in a patch update. I believe it should not release a new version at all.
> 
> Can you suggest another way to consume the updated packages than via a release?

I don’t think consumers should be impacted by internal changes of `devDependencies`

joecowton1:
@mxdvl because of the way packages are consumed by one another i still can't think of another way of doing this 😵‍💫 

@sndrs do you have any input on this one?

ob6160:
> I don’t think consumers should be impacted by internal changes of devDependencies, so they should not need to consume it.

I suppose this can be seen both ways. Looking at this through the lens of a consumer, I think I'd impacted by the internal changes here; in that if I didn't receive the updates included in this release I'd experience peer deps errors.

In the interests of not blocking peoples' time and based on my thinking above I'm of the opinion that the patch release is alright to go ahead with. We should discuss in more depth and agree an approach for this to prevent similar future scenarios.

ob6160 (approved):

PR #391, opened 2023-02-01
`loadScript` rejects with an Error

arelra:
## Background

The current most frequent error in Sentry, accounting for ~20% of DCR's errors is:

`Non-Error exception captured with keys: currentTarget, isTrusted, target, type`

This generic error is from Sentry complaining that it requires rejected promises to return a proper Error type:

https://github.com/getsentry/sentry-javascript/issues/2546#issuecomment-703331367

The errors look to be coming from `loadScript` given the serialised object reported by Sentry e.g.:

```js
{
  currentTarget: [object Null], 
  isTrusted: True, 
  target: head > script#ipsos[type="text/javascript"], 
  type: error
}
```

```js
{
  currentTarget: [object Null], 
  isTrusted: True, 
  target: head > script, 
  type: error
}
```

## This change

Updates promise rejections to return an Error type.

Changing the promise rejection won't reduce the errors but it will give us better insight into the errors reported by Sentry.

I suspect the underlying errors are the result of ad-blockers but we can't be sure without more detailed errors.

## Why?

To get better insight into high volume errors reported by Sentry

### DIFF
--- a/libs/@guardian/libs/src/loadScript/loadScript.test.ts
+++ b/libs/@guardian/libs/src/loadScript/loadScript.test.ts
@@ -75,6 +75,12 @@ describe('loadScript', () => {
 	});
 
 	it('rejects if the script fails to load', async () => {
-		await expect(loadScript(badURL)).rejects.toMatch(badURL);
+		expect.assertions(2);
+		try {
+			await loadScript(badURL);
+		} catch (error) {
+			expect(error instanceof Error).toEqual(true);
+			expect((error as Error).message).toEqual('Error loading script: bad-url');
+		}
 	});
 });
--- a/.changeset/thirty-pianos-occur.md
+++ b/.changeset/thirty-pianos-occur.md
@@ -1,0 +1,5 @@
+---
+'@guardian/libs': major
+---
+
+loadScript rejects with an Error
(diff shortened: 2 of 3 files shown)


libs/@guardian/libs/src/loadScript/loadScript.test.ts
```diff
@@ -75,6 +75,12 @@ describe('loadScript', () => {
...
 
 	it('rejects if the script fails to load', async () => {
-		await expect(loadScript(badURL)).rejects.toMatch(badURL);
+		expect.assertions(2);
+		try {
+			await loadScript(badURL);
+		} catch (error) {
+			expect(error instanceof Error).toEqual(true);
+			expect((error as Error).message).toEqual('Error loading script: bad-url');
+		}
```
sndrs, line 84: i think this makes it a breaking change? any code that's expecting a string if it fails will break?

arelra: Ah yes you're right I'll change it to a major release

arelra:
@ob6160 @joecowton1 

fyi the checks were stuck in pending status as a conversation was not resolved.

Once I resolved the conversation however the checks stayed stuck pending 😢 

I've done a no-edit commit and force pushed to kick it off again.

joecowton1:
> @ob6160 @joecowton1
> 
> fyi the checks were stuck in pending status as a conversation was not resolved.
> 
> Once I resolved the conversation however the checks stayed stuck in a pending state 😢
> 
> I've done a no-edit commit and force pushed to kick it off again.

Look ok now, let us know if its not!

sndrs (approved):

PR #399, opened 2023-02-02
Add more metadata to `loadScript` errors

arelra:
## What are you changing?

Small update to #391  to add more metadata to the error messages and to help distinguish them from each other.

I've seen that `targetSrc` can be undefined as I assume ad-blockers are removing them from the dom. However we still have reference to `src` from the closure so report that in all cases.


mxdvl (approved):
Isn’t this a `minor` as it’s a new feature?

mxdvl (approved):

PR #410, opened 2023-02-04
Fix a dead link for Locale

jirihofman:
## What are you changing?

- Link to Locale in `libs/@guardian/libs/README.md`.

## Why?

- The link is broken, `getLocale` folder doesn't exist.


ob6160 (approved):
good catch - thanks for the contribution!

PR #413, opened 2023-02-06
Update the ExpandingWrapper, Tab and FileInput styling according to the UX review

sophie-macmillan:
## What are you changing?
Expanding Wrapper
 - Multiline top border uses a light grey background
 - Changes the default collapsed height (and makes this configurable via a prop)
 - The Show more/less buttons are smaller
 - More/Less on the buttons are no longer capitalised
 
File Input
 - The default colour is neutral 7 
 
 Tabs
- Adds an indent so the last tab is indented to match the first tab

## Why?
Final UX review highlighted these changes
| before   |      After      |  
| ------------- | ------------- |
| image | image |
|image | image|
| image| image|

### DIFF
--- a/libs/@guardian/libs/README.md
+++ b/libs/@guardian/libs/README.md
@@ -22,7 +22,7 @@ Country data and methods to access it.
 
 Codified editorial design and information architecture.
 
-### [Locale](./src/getLocale)
+### [Locale](./src/locale)
 
 Get the user’s current location.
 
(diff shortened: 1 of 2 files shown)


.changeset/thick-tips-vanish.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/source-react-components-development-kitchen': patch
```
ob6160, line 2:
```suggestion
'@guardian/source-react-components-development-kitchen': minor
```

Is this a minor as it's introducing new non-breaking changes instead of being a bug fix (patch)?

ob6160 (approved):

PR #414, opened 2023-02-06
Add linkComponents to eslint settings

bryophyta:
## What are you changing?

Adding `Link` and `LinkButton` to the recommended eslint config for `source-react-components`.

## Why?

The `linkComponents` setting allows eslint rules to identify custom JSX elements which are being used as links.
For instance, adding this setting allows the
[`jsx-no-target-blank`](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md) rule to know that it should check both `<Link>` and `<LinkButton>` from Source components.

It felt like it might belong here (rather than in the DCR eslint config) because this is where the components are created and maintained (so the settings could be removed if the API for these components was changed). But I'd be happy to add it to DCR instead if the maintainers of source components would prefer.

### DIFF
--- a/.changeset/ninety-dogs-thank.md
+++ b/.changeset/ninety-dogs-thank.md
@@ -1,0 +1,9 @@
+---
+'@guardian/eslint-plugin-source-react-components': major
+---
+
+Add linkComponents to settings
+
+eslint allows us to specify components which should be treated as 'link components' for linting purposes.
+In source-react-components, both Link and ButtonLink will apply props directly to an `<a>` tag. Listing them as link components allows rules like `jsx-no-target-blank` to identify them as links, and to lint their props accordingly.
+If you don't have any linting rules which look for link elements then this update should not cause any changes. If you do have rules like this then it is possible that new errors or warnings will be raised, but these will be identifying pre-existing issues in your code.
--- a/libs/@guardian/eslint-plugin-source-react-components/src/index.ts
+++ b/libs/@guardian/eslint-plugin-source-react-components/src/index.ts
@@ -7,9 +7,15 @@ export const rules = {
 export const configs = {
 	recommended: {
 		plugins: ['@guardian/source-react-components'],
-
 		rules: {
 			'@guardian/source-react-components/valid-import-path': 'error',
+		},
+		settings: {
+			linkComponents: [
+				// Components used as alternatives to <a> for linking, eg. <Link to={ url } />
+				{ name: 'LinkButton', linkAttribute: 'href' },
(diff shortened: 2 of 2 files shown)


ob6160 (commented):
Great idea, seems like a really useful addition.

In terms of giving access to this configuration to users of the @source/react-components package, I'm wondering if this would be better placed in the `eslint-plugin-source-react-components` package: https://github.com/guardian/csnx/blob/main/libs/%40guardian/eslint-plugin-source-react-components/README.md

This is where we distribute rules for people who consume `@guardian/source-react-components`

joecowton1:
> Great idea, seems like a really useful addition.
> 
> In terms of giving access to this configuration for users of the `@guardian/source-react-components` package, I'm wondering if this would be better placed in the `@guardian/eslint-plugin-source-react-components` package: [`main`/libs/%40guardian/eslint-plugin-source-react-components/README.md](https://github.com/guardian/csnx/blob/main/libs/%40guardian/eslint-plugin-source-react-components/README.md?rgh-link-date=2023-02-07T10%3A37%3A39Z)
> 
> This is where we distribute rules for people who consume `@guardian/source-react-components`

Ditto what Ollie says here.

bryophyta:
okay thanks both! I'm pretty new to eslint so that's v helpful. I'll move it and push the change. Do either of you have a sense of what level this change would be? It should be a no-op in terms of the actual code being run, but could cause new errors to be flagged, depending on how the consumer's eslint is configured (e.g. if they do have the `jsx-no-target-blank` rule enabled).

ob6160:
> @ob6160 I've updated the branch following your suggestion. Any guidance on whether this should be considered major/minor/patch would be gratefully received :)

Hmm, I have a feeling that this might come under the "major" category as it adds new functionality that could break peoples' eslint build when they upgrade.

From [the docs(https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md#custom-link-components) it seems as though anyone who is using the `LinkButton` or `Link` component could potentially start to have a new set of errors due to the linter counting those as if they were `<a />` elements. @joecowton1 do you have any thoughts on this too?

I usually go by https://semver.org/ when I'm making decisions, it's very easy to get wrong by accident as the line can be blurry at times :)

ob6160 (approved):
nice one, thanks for the contribution! 💎

bryophyta:
Thanks for the review @ob6160 !

Just fyi I don't think I have the right permissions to merge in this repo, so someone else might have to do it when you have time, thanks!

PR #417, opened 2023-02-07
Add Telegram, Signal, WhatsApp brand icons

ob6160:
## What are you changing?

-

## Why?

-


libs/@guardian/source-react-components/scripts/create-icons/create-icon-component.ts
```diff
@@ -117,10 +117,12 @@ export const createIconComponent = async ({
 		},
 	);
 
-	const label = labels[icon.name];
+	const label = labels[icon.name] ?? icon.name;
 
 	if (!label) {
-		throw new Error(`No label found for ${icon.name}`);
+		console.warn(
```
joecowton1, line 123: Likewise here - do we need to change this?

ob6160: The icons in Figma don’t have labels so by throwing here the whole script stopped and I couldn’t complete the update.

Reasoning was that we could warn about the missing label and fall back to the icon id, so that the update could still go ahead

ob6160: (Manually added human readable labels after to the three new icons as I don’t have Figma edit access to add them at source)

libs/@guardian/source-react-components/scripts/create-icons/create-icon-component.ts
```diff
@@ -117,10 +117,12 @@ export const createIconComponent = async ({
 		},
 	);
 
-	const label = labels[icon.name];
+	const label = labels[icon.name] ?? icon.name;
```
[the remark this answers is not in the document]
mxdvl: `noUncheckedIndexedAccess` requires it to have a fallback?

mxdvl:
Worth noting that these SVGs have `id`s in them, meaning they can be added to a page only once without creating invalid HTML.

libs/@guardian/source-react-components/scripts/create-icons/create-icon-component.ts
joecowton1: Thanks for the explanation - is there a way that we can get someone with Figma edit access to add the labels at source?

joecowton1: Feels a bit like we might be changing the system here when the issue might be more about access to Figma

ob6160: Looking into why the label setup didn't work as-is in Figma. Agree that changing the system could be wrong as the enforcement here ensures that the right label is used.. In terms of this specific case, it does feel like the system is being quite inflexible by throwing. Debugging this and writing a fix yday to get the script to run cost me about 15-20 mins total, so I feel like capturing the change in some form might be of value to people who come to update icons in the future. How to capture it is an interesting question though :)

joecowton1: Yeah its an interesting point - is a comment to say "if this is throwing, check you have the correct label in Figma" too simple?

ob6160: yeah that's a good shout in this case I think - just doing some digging because it seems as though the label is there, but wasn't being picked up. Either way as this is seems to be an easy point of failure, the comment seems valuable. 

I'll back out the code change, leave the comment, and look into what's going wrong with the Figma integration.

This kind of integration seems really prone to this kind of breakage because of how flexible Figma is as a platform; so easy for someone to accidentally break the contract by moving a label outside of a frame or something.

libs/@guardian/source-react-components/scripts/create-icons/create-icon-component.ts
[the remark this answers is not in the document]
ob6160: `noUncheckedIndexedAccess` doesn't raise an error with this - so i've removed the null coalesce just to reduce the changes that this PR is making to the script logic. Will raise a ticket under the csnx gh project to look at fixing some of the problems raised here in the future

ob6160:
> Worth noting that these SVGs have `id`s in them, meaning they can be added to a page only once without creating invalid HTML.

thanks for pointing this out - i'll also add this to the ticket I'm going to put together to capture changes we want to make to fix/improve this script

libs/@guardian/source-react-components/scripts/create-icons/create-icon-component.ts
```diff
@@ -120,7 +120,11 @@ export const createIconComponent = async ({
 	const label = labels[icon.name];
 
 	if (!label) {
-		throw new Error(`No label found for ${icon.name}`);
+		// This error is thrown when the accessible label data for the icon is missing in Figma.
+		// If you need to generate the icon components quickly and fix the label later, you can comment out this line.
```
joecowton1, line 124: I think we might be wiser not suggesting this lest we set a precedent for bad habits - the comment on the line above tells people what they need to do.

ob6160: agreed - will remove

libs/@guardian/source-react-components/scripts/create-icons/create-icon-component.ts
joecowton1: Nice one, cheers.

PR #421, opened 2023-02-09
Export three icons: WhatsAppBrand, SignalBrand, and TelegramBrand

sophie-macmillan:
## What are you changing?
This exports the icons that were added in https://github.com/guardian/csnx/commit/93468258016dbdf78352f029fb10dea73af97f9c


ob6160:
@sophie-macmillan looks like there's a small change needed in the test too: libs/@guardian/source-react-components/src/index.test.ts

PR #424, opened 2023-02-09
Align ToggleSwitch perfectly

mxdvl:
## What are you changing?

- `ToggleSwitch` alignment
   image

- Extract out the dynamic part of styles

## Why?

- Things are more precisely aligned
- Styles are more performant when more static

### DIFF
--- a/libs/@guardian/source-react-components-development-kitchen/src/toggle-switch/styles.ts
+++ b/libs/@guardian/source-react-components-development-kitchen/src/toggle-switch/styles.ts
@@ -71,7 +71,7 @@ export const toggleStyles = (format?: ArticleFormat): SerializedStyles => {
 			top: 5px;
 			height: 11px;
 			width: 6px;
-			right: 10px;
+			right: 8px;
 			opacity: 0;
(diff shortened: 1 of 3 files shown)


libs/@guardian/source-react-components-development-kitchen/src/toggle-switch/styles.ts
```diff
@@ -24,12 +24,10 @@ const toggleBorder = 'rgba(255, 255, 255, 0.6)';
  */
 const toggleBorderGreen = '#A7CFB8';
 
-export const buttonStyles = (
```
OllysCoding, line 27: Are these only used by ToggleSwitch?

OllysCoding (approved):
+1 I am visually satisfied :)

libs/@guardian/source-react-components-development-kitchen/src/toggle-switch/styles.ts
mxdvl: They’re inside the toggle-switch folder and typescript does not complain with the signature change and no visual regression, so seems safe to me!

PR #453, opened 2023-02-17
Update form field borders

abeddow91:
## What are you changing?

Reducing the border width on form fields to 1px and updating their border radius to 4px. Form fields refer to Text input, Text area, Radio button, Checkbox, Select box, Choice card, and User Feedback.

 
## Why?

These changes reflect styling updates made to these components by the Design System Team.

Their reason is the following:

"The existing form fields had a much thicker border. The impact of this was that the fields dominated the page and affected the visual hierarchy. That’s why we have decided to make field borders 1px across the board and update corner radius in all instances."

### DIFF
--- a/.changeset/brown-hounds-rhyme.md
+++ b/.changeset/brown-hounds-rhyme.md
@@ -1,0 +1,5 @@
+---
+'@guardian/source-react-components': minor
+---
+
+update border width and radius on form field components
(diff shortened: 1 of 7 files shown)


joecowton1:
Looks great, thanks Anna!

joecowton1 (approved):

PR #457, opened 2023-02-22
docs: use JSDOC to indicate deprecated exports

cemms1:
## What are you changing?

- Change the comments above the deprecated palette exports to JSDOC comments

## Why?

- Gives more immediate feedback to consumers in IDEs that the exports are deprecated which encourages users to import using other non-deprecated methods.


joecowton1 (approved):
Great idea.

PR #464, opened 2023-02-28
add optional param to checkboxGroup and radioGroup

marjisound:
## What are you changing?
This PR adds `optional` param to the `CheckboxGroup` & `RadioGroup` elements
-

## Why?
These elements need to have the `Optional` next to their title if the consumer indicates that they are optional

| Before      |    After      | 
|-------------|------------|
| image | image | 
| image | image |

### DIFF
--- a/libs/@guardian/source-react-components/src/radio/RadioGroup.tsx
+++ b/libs/@guardian/source-react-components/src/radio/RadioGroup.tsx
@@ -25,6 +25,10 @@ export interface RadioGroupProps
 	 */
 	hideLabel?: boolean;
 	/**
+	 * Adds the word "Optional" after the label
+	 */
+	optional?: boolean;
+	/**
 	 * Additional text or component that appears below the label
 	 */
 	supporting?: string | JSX.Element;
@@ -52,6 +56,7 @@ export const RadioGroup = ({
 	id,
 	name,
 	label,
+	optional = false,
 	hideLabel = false,
 	supporting,
 	orientation = 'vertical',
@@ -61,8 +66,15 @@ export const RadioGroup = ({
 	...props
 }: RadioGroupProps): EmotionJSX.Element => {
 	const groupId = id ?? generateSourceId();
+	console.log(`optional: ${optional.toString()}`);
+
 	const legend = label ? (
-		<Legend text={label} supporting={supporting} hideLabel={hideLabel} />
+		<Legend
+			text={label}
+			supporting={supporting}
+			hideLabel={hideLabel}
+			optional={optional}
+		/>
 	) : (
 		''
 	);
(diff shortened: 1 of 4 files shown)


marjisound:
[Optional radio fields are not showing the optional text](https://trello.com/c/HvBrsXDX/812-optional-radio-fields-are-not-showing-the-optional-text)

libs/@guardian/source-react-components/src/radio/RadioGroup.tsx
```diff
@@ -61,8 +66,15 @@ export const RadioGroup = ({
 	...props
 }: RadioGroupProps): EmotionJSX.Element => {
 	const groupId = id ?? generateSourceId();
+	console.log(`optional: ${optional.toString()}`);
```
joecowton1, line 69: Left a console.log in here.

libs/@guardian/source-react-components/src/radio/RadioGroup.tsx
```diff
@@ -91,19 +103,21 @@ export const RadioGroup = ({
...
-		>
-			{legend}
-			{message}
-			{orientation === 'vertical' ? (
-				<Stack>{radioContainers}</Stack>
-			) : (
-				<Inline space={5}>{radioContainers}</Inline>
-			)}
-		</fieldset>
+		<>
```
joecowton1, line 106: Don't think we need these fragments

joecowton1 (commented):
Thanks Marjan. A couple of minor things but looks good otherwise!

libs/@guardian/source-react-components/src/radio/RadioGroup.tsx
marjisound: Ooops sorry, just deleted it

libs/@guardian/source-react-components/src/radio/RadioGroup.tsx
marjisound: My bad, just removed it 👍

joecowton1 (approved):
LGTM. We should let @akemitakagi approve before merging.

mxdvl:
What does an optional radio button mean?

- https://www.sarasoueidan.com/blog/one-checkbox-or-two-radio-buttons/
- https://www.nngroup.com/articles/radio-buttons-default-selection/

akemitakagi:
Hi @marjisound 
Thanks for adding the optional param to the radio button and checkboxes.
We have them in the Figma component library. I didn't realise we don't have these variations in the code.

akemitakagi:
Hi @mxdvl 
It's an optional field in a form that users only need to answer if they want to. 
Example.
Screenshot 2023-02-28 at 12 38 47

akemitakagi (approved):

mxdvl:
> Hi @mxdvl It's an optional field in a form that users only need to answer if they want to.

As far as I am aware, it’s not possible to unselect a radio button, so it’s only ever temporarily optional. The article I mentioned go into more depth, but [even the MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio#data_representation_of_a_radio_group) suggest that it’s a pattern worth thinking about:

> It's fairly uncommon to actually want to allow the form to be submitted without any of the radio buttons in a group selected, so it is usually wise to have one default to the checked state. See [Selecting a radio button by default](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio#selecting_a_radio_button_by_default) below.

akemitakagi:
@mxdvl Thanks for this!
We will review our usage of the optional radio buttons.

marjisound:
Thanks @mxdvl  for pointing this out. Yea, with the current setup it's not possible to unselect a radio button after it's been selected.  

@akemitakagi would you suggest holding off this PR to make sure what approach we want to take regarding the unselect?

mxdvl:
Further reading from the Government Design System:
- https://design-system.service.gov.uk/components/checkboxes/
- https://design-system.service.gov.uk/components/radios/
  > Users cannot go back to having no option selected once they have selected one, without refreshing their browser window. Therefore, you should include ‘None of the above’ or ‘I do not know’ if they are valid options.

marjisound:
No need for this PR: 
The DS and UX teams decided to use radio button groups as a mandatory field.
The guidance is that we include a 'none of the above' option if needed so that the reader can select one of these options.

PR #465, opened 2023-02-28
feat: add a User-Agent parser

mxdvl:
## What are you changing?

- This formalises the process of normalising our browser in our Data Warehouse.

## Why?

- We want to be able to update our existing `browserlist` without relying on third-party data sources.


apps/ua-parsing/src/main.js
```diff
@@ -0,0 +1,124 @@
+import { UAParser } from './ua-parser.js';
```
sndrs, line 1: why not use `https://www.npmjs.com/package/ua-parser-js`?

mxdvl: We’re actually using 99% of the source code of that package, [but its current implementation has very outaded special exports to handle cases where it’s invoked from jQuery](https://github.com/faisalman/ua-parser-js/blob/master/src/ua-parser.js#L1132-L1153), so it was easier to lift the internal part to generate code that could run inside BigQuery.

PR #466, opened 2023-03-01
Marji/add optional to checkbox

marjisound:
## What are you changing?

This PR adds `optional` param to the `CheckboxGroup` element

## Why?

This element needs to have the `Optional` next to their title if the consumer indicates that they are optional

| Before      |    After      | 
|-------------|------------|
| image | image |

### DIFF
--- a/.changeset/metal-days-sit.md
+++ b/.changeset/metal-days-sit.md
@@ -1,0 +1,5 @@
+---
+'@guardian/source-react-components': minor
+---
+
+Adds optional prop to CheckboxGroup element
--- a/libs/@guardian/source-react-components/src/checkbox/CheckboxGroup.stories.tsx
+++ b/libs/@guardian/source-react-components/src/checkbox/CheckboxGroup.stories.tsx
@@ -75,6 +75,13 @@ SupportingTextDefaultTheme.args = {
 
 // *****************************************************************************
 
+export const OptionalDefaultTheme = Template.bind({});
+OptionalDefaultTheme.args = {
+	optional: true,
+};
+
+// *****************************************************************************
+
 export const SupportingTextBrandTheme = Template.bind({});
 SupportingTextBrandTheme.parameters = {
 	backgrounds: {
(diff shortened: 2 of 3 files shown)


marjisound:
This change was originaly within https://github.com/guardian/csnx/pull/464 PR but because that PR is on hold, awaiting for some discussion about what to do with the optional radio group since with an optional radio group, when the user selects one radio button, they wouldn't be able to de-select it afterwards.

.changeset/metal-days-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/source-react-components': minor
+---
+
+Adds optional prop to CheckboxGroup element
```
joecowton1, line 5:
```suggestion
Adds optional prop to CheckboxGroup element to append "optional" text to label
```

joecowton1 (approved):

marjisound:
[Optional radio fields are not showing the optional text](https://trello.com/c/HvBrsXDX/812-optional-radio-fields-are-not-showing-the-optional-text)

mxdvl:
Is a checkbox not optional by default?

What’s the status of an unchecked, but not optional checkbox?

marjisound:
> Is a checkbox not optional by default?
> 
> What’s the the expected behaviour of an unchecked, but not optional checkbox? Prevent form submission?

This is just for guiding the user visually. In the form we show the word `optional` next to the label of all the optional fields apart from checkbox group. 
Although now that I'm thinking, the checkbox can never be mandatory, because if the user doesn't check it, it would go with value false. Let me talk to our UX designer and confirm this.

marjisound:
@mxdvl The design system team said there should be an optional label present if the checkbox group question is optional, and hide the label if it's mandatory (and show error on form submission if empty)

This is because sometimes these checkboxes are used for multiple choice questions and could be mandatory where the user needs to check at least one of them. But they can also check more than one if they want.

marjisound:
Hi @mxdvl I may not have fully understood what you mean, but here the scenario is when we have a group of checkboxes and we want the user to at least check one of them. So each individual checkbox is not required on its own, but choosing at least one of them is required. What do you think?

mxdvl:
As I've said, I can't think of a good use case for this and this am a little concerned we are allowing for confusing patterns in our forms.

I've added it as an agenda for the next Accessibility Champions meetup, as I think a synchronous chat would be helpful in this instance. Hopefully my worries are unfounded.

PR #475, opened 2023-03-09
Update `@guardian/browserlist-config` data

sndrs:
## What are you changing?

- adding latest browser data to `@guardian/browserlist-config`

## Why?

- in the name of accuracy


arelra (approved):
Been eagerly waiting for Safari 10 support to drop below 0.01%

Good news:
Safari 10.1 has dropped below 0.01%

Not so good news:
Safari mobile 10.3 increased from 0.05 to 0.06% (suprisingly)

PR #473, opened 2023-03-07
Refactor chromatic workflows

sndrs:
## What are you changing?

- move github pages storybook deployment to its own workflow (triggered by successful CI workflow run, only on main)
- run the chromatic tests in a matrix

## Why?

- simplifies the workflow files
- the chromatic jobs can run in parallel

## Before 
`3m 23s` (made longer by the number of chromatic runs)

image

## After
`2m 59s` (only made longer by the longest chromatic run)

image

PR #478, opened 2023-03-13
Update form field borders styles

abeddow91:
## What are you changing?
This PR updates the borders across multiple states on multiple form components listed below. These state changes are:

- Default state: 1px
- Hover state: 
    1px for Text input, Text area, and Select box. (No colour change. The cursor style change)
    2px for Radio button, Checkbox, and Choice card.
- Selected state: 2px (Radio button, Checkbox, and Choice card.)
- Error state: 2px
- Success state: 2px
- Informative state: 2px User Feedback
- Focus state: 3px outside of default, error, and success border.

The affected components are: 

- Radio button
- Checkbox
- Text Input
- Text Area
- Select Box
- Choice card
- User Feedback summary

**visual changes can be viewed by running Storybook or in chromatic diffs**

## Why?

These changes reflect changes requested by the Editorial Design team.


.changeset/spotty-garlics-retire.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/source-react-components': minor
+---
+
+Update border styles across form components
```
ob6160, line 5: Looks like we have two changesets in here

ob6160 (commented):
nice work - I'm curious as to why we can't see the Chromatic diff in the GitHub workflow summary (I'd expect it to show that there have been changes to stories). this could be related to the changes in https://github.com/guardian/csnx/pull/473

joecowton1:
Looks like some of the UI changes are showing in Chromatic now @abeddow91. 

If there are changes to dev kitchen (and source foundations too?) too do we need add changesets for those too?

joecowton1 (approved):
Nice work 👌

mxdvl:
I wonder if such a change should better be represented by a major version. Could this potentially break consumers who are not using `box-sizing: border-box;`?

abeddow91:
> I wonder if such a change should better be represented by a major version. Could this potentially break consumers who are not using `box-sizing: border-box;`?

@mxdvl Its a good point! Technically I suppose that would be a breaking change in that circumstance. I suspect that we rarely have consumers that are not using `box-sizing: border-box;` but we can't be sure and as @sndrs pointed out, the [docs specify visual changes are major]( https://github.com/guardian/csnx/blob/main/docs/source/contributing.md#versioning). I'll update the changesets but I wonder if this raises a bigger question around how we version source - it feels like a lot to release a major change for something that would affect so few consumers.

PR #479, opened 2023-03-13
Ensure chromatic uses the correct tokens

sndrs:
## What are you changing?

- use the correct chromatic tokens in the chromatic runs

## Why?

- in #473 i missed the token, which i think means it's not picking up changes in #478

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,13 +35,15 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


ob6160 (approved):
thanks for looking into this!

joecowton1:
Thanks @sndrs - I've removed the branch checks so it should pass now, might need re-running though.

joecowton1:
Oddly the atoms-rendering UI tests / UI review aren't showing in the list of checks.

PR #490, opened 2023-03-17
remove default jsx handler from `tsconfig.base.json`

sndrs:
## What are you changing?

- make all project define how they handle JSX

## Why?

- defining a default creates issues when you don't know about it. so require explicit handling


mxdvl (approved):
This seems very sound!

PR #497, opened 2023-04-11
Types via JSDoc for @guardian/libs

mxdvl:
## What are you changing?

- Convert `@guardian/libs` to pure JavaScript
- Rely on [JSDoc comments for type-checking](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html)

## Why?

- Reduce the burden of using TypeScript for our library code.
- Faster builds
- Easier to introspect: no source maps, no declaration file, copy-and-paste anywhere
- Consumers will bundle the code again, [we already target ES2020](https://github.com/guardian/csnx/blob/96dad00147b631d22095dd70d48b69ecc55aad88/libs/%40guardian/libs/README.md#bundling) / [ES2018](https://github.com/guardian/csnx/blob/96dad00147b631d22095dd70d48b69ecc55aad88/tools/nx-plugins/npm-package/build/index.ts#L44-L47)

## Discussion

We should consider updating our [recommendations for NPM packages](https://github.com/guardian/recommendations/blob/main/npm-packages.md#compiling)


libs/@guardian/libs/src/logger/log.js
```diff
@@ -1,47 +1,43 @@
...
+import { isString } from '../isString/isString.js';
+import { storage } from '../storage/storage.js';
+import { STORAGE_KEY } from './storage-key.js';
+import { commonStyle, isTeam, teamStyles } from './teamStyles.js';
 
 const allStyles = { ...teamStyles, ...commonStyle };
 
-const messageStyle = (teamStyle: TeamStyle): string => {
+/** @typedef {import('./teamStyles').TeamName} TeamName */
+/** @typedef {typeof subscribeTo} TeamSubscription */
```
mxdvl, line 9: inline types

libs/@guardian/libs/tsconfig.json
```diff
@@ -1,5 +1,8 @@
 {
 	"extends": "../../../tsconfig.base.json",
+	"compilerOptions": {
+		"checkJs": true
+	},
```
mxdvl, line 5: We want to make sure our JavaScript is type-checked.

libs/@guardian/libs/src/loadScript/loadScript.js
```diff
@@ -44,11 +35,12 @@ export const loadScript = (
 			}
 
 			if (event instanceof Event) {
-				const target = event.target as Element;
-				const targetSrc = target.getAttribute('src') ?? '';
+				const target = event.target;
+				const targetSrc =
+					target instanceof Element ? target.getAttribute('src') : null;
```
mxdvl, line 40: Remove type assertions, so `targetSrc` is now `string | null`

libs/@guardian/libs/src/loadScript/loadScript.js
```diff
@@ -58,5 +50,5 @@ export const loadScript = (
 		};
 
 		const ref = document.scripts[0];
-		ref?.parentNode?.insertBefore(script, ref);
+		if (ref) ref.parentNode?.insertBefore(script, ref);
```
mxdvl, line 53: Small refactor to help the compiler understand that `ref` is defined

mxdvl (commented):
Pointer to some of the changes.

- `as const` assertion [uses JSDoc casts](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#casts)
- external types are inlined with [`@typedef`](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#typedef-callback-and-param)
- `@guardian/libs` no longer uses the TS Rollup plugin
- Ophan types are left untouched
- minor refactors are pointed to below
- `ArticleFormat` does not uses [`@enum`](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#enum) as the type-cast are more explicit

libs/@guardian/libs/src/format/ArticleDesign.js
```diff
@@ -0,0 +1,31 @@
+/** @typedef {typeof ArticleDesign[keyof typeof ArticleDesign]} ArticleDesign */
+
+export const ArticleDesign =
+	/** @type {const} @satisfies {Record<string, number>} */ ({
+		Standard: 100,
+		Gallery: 101,
```
mxdvl, line 6: Why start at `100`?

To prevent clashes with `ArticleTheme` (0-7) and `ArticleDisplay` (200-203)

PR #498, opened 2023-04-11
Add review_requested to Chromatic GH action

joecowton1:
## What are you changing?

- adding a `review_requested` event type and only triggering chromatic if true

## Why?

- we're trialling minimising the number of snapshots we create


ob6160 (approved):
lgtm!

PR #504, opened 2023-04-12
Revert use of react jsx runtime instead of @emotion/react jsx runtime (#504

joecowton1:
## What are you changing?

- reverting https://github.com/guardian/csnx/pull/489 

## Why?

We're seeing a strange bug where typescript errors in projects when you build a project that is not configured to handle JSX, even if that project contains no jsx itself. This seems related to imports in the dist folder. More details below.

We haven't been able to fix this bug and while CI on csnx passes due to caching, we don't want a footgun like this floating around in the background.  

***

From google chat thread:

Alex Sanders:

After thrashing this out with Mahesh Makani and Joe Cowton all afternoon and some further prodding, the issue seems to arise when:
- you build a project that is not configured to handle JSX 
- Nx doesn't have a cache result for that project (or you skip the cache)
- something in dist has a .d.ts that imports from @guardian/react-components (i.e. dev kitchen)

if you remove all import [whatevers] from "@guardian/source-react-components"; from dist/**/*.d.ts then everything seems fine hence this error

TS has found a declaration file that imports from a package that the paths config resolves to the react components source directory, and because the project is not set up handle jsx, it errors

what i don't know is why TS is worried about declaration files in `dist` at any point, let alone when building an unrelated project

### DIFF
--- a/.changeset/thick-dancers-flash.md
+++ b/.changeset/thick-dancers-flash.md
@@ -1,0 +1,5 @@
+---
+'@guardian/ab-react': patch
+---
+
+Use `@emotion/react` jsx runtime instead of `react` jsx runtime
--- a/libs/@guardian/ab-react/package.json
+++ b/libs/@guardian/ab-react/package.json
@@ -6,6 +6,7 @@
 	"license": "Apache-2.0",
 	"sideEffects": false,
 	"devDependencies": {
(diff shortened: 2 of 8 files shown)


mxdvl (approved):
Bit of a shame, but at least the builds are deterministic!

PR #510, opened 2023-04-18
Group radio buttons in `Answers` story so they are selectable

jamesmockett:
## What are you changing?

- Wraps the radio buttons in the `<Answers>` story with a `<RadioGroup>` so they are selectable.

## Why?

- The quiz radio buttons override the default Source radio button styles. Whilst reviewing changes to the form element border styles (#478) it was discovered that the examples in these stories were not selectable so that the different interaction states could not be tested.


ob6160 (approved):
lgtm!

PR #511, opened 2023-04-18
Platform | Run Chromatic on main 

ob6160:
## What does this change?
- Right now the Chromatic job is skipped on the main branch because the conditional check for the `run_chromatic` label fails.
- This extends that conditional check to also be truthy if the branch is `main`
- Follow-on from #498


jamesmockett (approved):
Good spot 👍

PR #513, opened 2023-04-20
Revert manually updated package versions and change logs

jamesmockett:
## What are you changing?

- Reverting the manual package version bumps and change log updates introduced in #478

## Why?

- This was missed during review, but these updates should be handled via a changeset. With the manual changes in place the changeset bumps the package versions again and duplicates the information in each package's changelog.


jamesmockett:
Closing in favour of reverting the original PR: #514

PR #516, opened 2023-04-21
Rerun Chromatic when changes made to PR

jamesmockett:
## What are you changing?

- Reruns Chromatic when changes are made to a PR if `run_chromatic` label is present.

## Why?

- With the existing workflow (#498) Chromatic is only triggered when the label is first added. Subsequent pushes _do not_ trigger Chromatic without removing and adding the label again which adds friction.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on:
     branches:
       - main
   pull_request:
-    types: [labeled, synchronize]
+    types: [opened, labeled, synchronize]
 
 jobs:
   validate:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -4,7 +4,7 @@ on:
     branches:
       - main
   pull_request:
-    types: [labeled, synchronize]
+    types: [opened, labeled, synchronize]
```
joecowton1, line 7: Do we still need `synchronize` as a type here?

jamesmockett: I thought we did, but perhaps not? 🤔 Without it this action should still run once a PR has been opened. I'll trying removing it…

jamesmockett: No, it is still needed. Without `synchronize` pushes to the branch no longer trigger the action.

joecowton1: ok great - thanks for checking!

joecowton1 (approved):

PR #517, opened 2023-04-24
Update `create-icons` accessible label list

jamesmockett:
## What are you changing?

- Updates the `create-icons` script to include accessible labels for newly added icons.
- Updates the README with the correct command to run the script and steps to obtain a Figma access token.

## Why?

- Currently it's not obvious how you run the script.
- If you do run the script it will fail due to an incomplete list of icons.

Ultimately we want to fetch the accessible name from Figma rather than having to maintain a lookup, but this isn't straightforward due to how the document is currently structured.


joecowton1:
Nice work - we probably need to add a changeset?

joecowton1:
> > Nice work - we probably need to add a changeset?
> 
> Does this need a changeset as it's a tooling update and doesn't include any consumer facing changes?

I never know for certain but I was going off the following:
- update to the readme
- fix to the script

I feel like either of those could justify a patch, that way the update is also recorded in the changelog

joecowton1 (approved):
Nice one!

PR #520, opened 2023-04-25
Update testing-library packages

joecowton1:
## What are you changing?

- updating `testing-library` packages

## Why?

- keeping things up to date.
- we need to update the nrwl/jest package which is proving tricky, doing this first eliminates some variables


jamesmockett (approved):
Good idea to upgrade this in isolation first 👍

PR #521, opened 2023-04-25
Apply focus state management to `atoms-rendering` stories

jamesmockett:
## What are you changing?

- Adds decorator to apply focus state management to `atoms-rendering` stories.

## Why?

- As part of the design review for #478 it was noted that the keyboard focus states were being triggered when clicking on form elements.

### Before

Screenshot 2023-04-25 at 10 33 02

### After

Screenshot 2023-04-25 at 10 33 30


joecowton1:
LGTM, I think we want to make sure that this isn't also a problem in production (ie when Source is used in DCR).

joecowton1 (approved):

jamesmockett:
> LGTM, I think we want to make sure that this isn't also a problem in production (ie when Source is used in DCR).

This is an example of the component in the wild on DCR and focus state is being correctly managed:
https://www.theguardian.com/games/2022/dec/11/quiz-of-the-year-how-well-do-you-remember-2022

PR #523, opened 2023-04-25
Upgrade NX packages to latest versions

joecowton1:
## What are you changing?

Upgrading all nrwl packages. This involves:

- removing workspace-lint (deprecated)
- deleting the workspace.json (deprecated)
- adding project names 
- adding `@testing-library/jest-dom` and `@types/youtube` to atoms-rendering `types` in the tsconfig
- overriding the `transform` field of jest.config for projects without TS.

PR #524, opened 2023-04-25
Move atoms-rendering window types to root

joecowton1:
## What are you changing?

- move window types for atoms-rendering to the window.d.ts in the root @types folder
- add conditionals for `window.google` to youtube atom

## Why?

- Having the `window` definitions in a tsx file was causing issues when upgrading to the latest nx packages in https://github.com/guardian/csnx/pull/523


libs/@guardian/atoms-rendering/src/YoutubeAtomPlayer.tsx
```diff
@@ -19,7 +19,7 @@ import type { AdTargeting, VideoEventKey } from './types';
 import type { PlayerListenerName } from './YouTubePlayer';
 import { YouTubePlayer } from './YouTubePlayer';
 
-declare class ImaManager {
+export declare class ImaManager {
```
sndrs, line 22: i think this should not use `declare`, to stop it being ambient itself

```suggestion
export class ImaManager {
```

@types/window.d.ts
```diff
@@ -1,5 +1,6 @@
 import type { TeamSubscription } from '../libs/@guardian/libs/src/logger/@types/logger';
 import type { Switches } from '../libs/@guardian/libs/src/switches/@types/Switches';
+import { ImaManager } from 'libs/@guardian/atoms-rendering/src/YoutubeAtomPlayer';
```
sndrs, line 3: this path is wrong, which makes me thing the `declare` below was still making it ambient (assuming nothing errored?)

```suggestion
import { ImaManager } from '../libs/@guardian/atoms-rendering/src/YoutubeAtomPlayer';
```

@types/window.d.ts
```diff
@@ -38,5 +39,13 @@ declare global {
 				pageViewId: string;
 			};
 		};
+		/**
+		 * Here we want to type the google object that will be added to window.
+		 * Since the imported google namespace is a value rather than a type, use typeof.
+		 */
+		google: typeof google;
```
sndrs, line 46: where does the `google` type come from?

mxdvl: (used to come from https://github.com/guardian/atoms-rendering/blob/6b4a78543898acf5fde44203b2b5a6ef693daec0/src/ima.d.ts)

libs/@guardian/atoms-rendering/src/YoutubeAtomPlayer.tsx
joecowton1: as in leave it as it was?

joecowton1: ah ok great!

arelra (approved):
lgtm 👍

PR #525, opened 2023-04-26
Use @types/react v18.2.0

joecowton1:
## What are you changing?

- updating the version of @types/react we use inline with our React version


mxdvl (approved):
Wunderbar

PR #540, opened 2023-05-11
Update nx and related deps

joecowton1:
## What are you changing?

- Updating nx and related packages

## Why?

- we're upto v16


nx.json
```diff
@@ -54,28 +51,13 @@
 			}
 		}
 	},
-	"targetDependencies": {
-		"build": [
-			{
-				"target": "build",
-				"projects": "dependencies"
-			}
+	"namedInputs": {
```
ob6160, line 54: what's this new part doing?

ob6160 (approved):
looks good, one question

nx.json
joecowton1: Good question, `namedInputs` replaces `implicitDependencies` in nx 16: https://nx.dev/deprecated/global-implicit-dependencies

PR #573, opened 2023-05-17
Add Jest as common dependency to workspace root

### LINKED ISSUE #572: Add Jest as dependency to root `package.json`
Jest is an explicit dependency in the `@csnx/hello-world` project only and appears to be being hoisted when running tests within other projects. We should either explicitly add Jest as a dependency to each project or to the root `package.json`.

jamesmockett:
Resolves #572

## What are you changing?

- Removes Jest and associated dependencies from `@csnx/hello-world` project and adds them to the workspace root.

## Why?

- Currently Jest is only installed as a dependency in `@csnx/hello-world`, but is being used by other projects in the monorepo when running tests. We could either add Jest as a dependency to each project that requires it or—as has been done here—as a common dependency to the workspace root.


joecowton1 (approved):
Nice work!

Out of curiosity: do we not need @types/jest anymore then?

jamesmockett:
> Out of curiosity: do we not need @types/jest anymore then?

`@types/jest` was already there in the root `package.json` so it's not shown in the diff as I didn't need to add it. (But it is still there 😄)

PR #574, opened 2023-05-19
Remove --all flag

joecowton1:
## What are you changing?

- removing `--all-true` from nx make scripts			

## Why?

- it's not advised on large repos (and is [deprecated](https://nx.dev/packages/nx/documents/run-many#all) in its current syntax)


sndrs (commented):
why not?

PR #575, opened 2023-05-19
Give write permissions to changesets and storybook actions

joecowton1:
## What are you changing?

- adding `contents: write` and `pull_request: write` permissions for storybook and changesets workflows 

## Why?

- both jobs are currently failing in CI due to lack of permissions

### DIFF
--- a/.github/workflows/changesets.yml
+++ b/.github/workflows/changesets.yml
@@ -7,6 +7,7 @@ on:
 
 permissions:
   contents: write
+  pull-requests: write
 
 env:
   NX_BRANCH: ${{ github.event.pull_request.head.ref }}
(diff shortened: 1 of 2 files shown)


mxdvl (commented):
I would consider adding the permissions [to individual jobs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idpermissions) instead.

~I’m also confused as to whether you would need both `read` and `write`, and could not find a definitive answer in the docs.~ EDIT: [`write` implies `read`](https://github.com/orgs/community/discussions/26655#discussioncomment-3252745) h/t @joecowton1

PR #577, opened 2023-05-19
Skip Chromatic in Dependabot PRs

### LINKED ISSUE #570: CSNX: Optimise Chromatic on dependabot PR's
Can we optimise the workflow around dependabot PR's so we don't have to manually add the `run_chromatic` label each time?

The DCR Chromatic actions might be a useful prompt here.

jamesmockett:
## What are you changing?

- Runs Chromatic against Dependabot PRs with the `skip` flag enabled.
- Resolves #570

## Why?

- To cut down on the number of Chromatic snapshots we don't want to run Chromatic against Dependabot PRs as they tend to hang around for a while before being merged and get rebased multiple causing unnecessary runs. As the UI tests are a required merge check we run Chromatic with the `skip` flag which marks these as passing and does not prevent merging of the PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,10 +29,14 @@ jobs:
       - run: make build-storybooks
 
   chromatic:
-    # Kick chromatic off only once all other steps have passed. This stops us
+    # Kick Chromatic off only once all other steps have passed. This stops us
     # wasting money running checks on PRs that are going to fail anyway.
+    # Chromatic is only run if the PR is labelled with `run_chromatic`, or we're
+    # on the `main` branch.
     needs: validate
-    if: contains(github.event.pull_request.labels.*.name, 'run_chromatic') ||  github.ref == 'refs/heads/main'
+    if: |
+      (contains(github.event.pull_request.labels.*.name, 'run_chromatic') || github.ref == 'refs/heads/main') &&
+      github.actor != 'dependabot[bot]'
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


joecowton1 (commented):
This is great James! I wondered if rather than repeating all the code we could just do something like:
```
skip: github.actor == 'dependabot[bot]'
```
would that work?

joecowton1:
> > This is great James! I wondered if rather than repeating all the code we could just do something like:
> > ```
> > skip: github.actor == 'dependabot[bot]'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > would that work?
> 
> Ooo, I hadn't thought of that. That would be a lot simpler! 😅 I'll give it a go…

Thanks! I've no idea if that actually would work but maybe something similar either way

.github/workflows/ci.yml
```diff
@@ -71,4 +78,4 @@ jobs:
           storybookBuildDir: 'dist/storybook/libs/@guardian/${{ matrix.lib }}'
           exitOnceUploaded: true
           onlyChanged: '!(main)' # turbosnap on non-main branches
-
+          skip: ${{github.actor != 'dependabot[bot]'}}
```
joecowton1, line 81: You'd need this the other way around i think?

.github/workflows/ci.yml
```diff
@@ -29,10 +29,17 @@ jobs:
...
+    # Chromatic is only run if the PR is labelled with `run_chromatic`, this is
+    # the `main` branch, or this is a Dependabot PR. For Dependabot PRs we then
+    # apply the `skip` flag as we don't want to create snapshots, but this
+    # ensures the UI tests are marked as passing allowing the PR to be merged.
     needs: validate
-    if: contains(github.event.pull_request.labels.*.name, 'run_chromatic') ||  github.ref == 'refs/heads/main'
+    if: |
+      (contains(github.event.pull_request.labels.*.name, 'run_chromatic') ||
+      github.ref == 'refs/heads/main') ||
+      github.actor != 'dependabot[bot]'
```
joecowton1, line 42: And to remove this?

joecowton1 (approved):

PR #579, opened 2023-05-22
Update svgr packages

joecowton1:
## What are you changing?

- updating svgr dev dependancies		

## Why?

- housekeeping


jamesmockett:
Do we still need `@types/svgo`? SVGO 3.x includes type definitions so seems like these aren't needed:
https://www.npmjs.com/package/@types/svgo

PR #588, opened 2023-05-24
`IdentityAuth` implementation in `@guardian/identity-auth`

coldlink:
## What are you changing?

Adding a library for authenticating and authorising users within the Guardian's identity platform hosted by Okta using OAuth2 and OpenID Connect within a client-side, browser-based context.

Hoping for it to be reviewed by CSTI, Identity, and Web Ex!

## Why?

This is a slimmed down and reverse engineered version of [Okta's official JS SDK](https://github.com/okta/okta-auth-js) that is tailored to the Guardian's needs.

Specifically, it provides a subset of the functionality of the official SDK. This is to reduce the size of the library and to ensure that we only expose the functionality that we need. Our library only uses some of the same classes and methods as the official SDK, but the implementation of them is our own.

Not sure how accurate this VSCode plugin is, but it gives an idea!

Screenshot 2023-06-07 at 10 18 08

**Webpack + SWC + Modern browsers**
_Okta Library_
Screenshot 2023-06-07 at 10 18 32

_This PR/IdentityAuth_
Screenshot 2023-06-07 at 10 18 49

# How to use

## Browser support

Browsers will need to support the following features or polyfill them:

- [`Web Crypto API`](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API)
  - [`crypto.getRandomValues`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues)
  - [`crypto.subtle.digest`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest)
  - [`crypto.subtle.importKey`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/importKey)
  - [`crypto.subtle.verify`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/verify)
- [`TextEncoder`](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder)

## Usage

```js
import { IdentityAuth } from '@guardian/identity-auth';

// Initialise the library with your configuration values
const identityAuth = new IdentityAuth({
	issuer: 'https://profile.theguardian.com/oauth2/SERVER_ID',
	clientId: 'example-client-id',
	redirectUri: 'https://theguardian.com',
	scopes: ['openid', 'profile'], // and any other scopes you need
});

// Check if the user is logged in and return the current auth state
const authState = await identityAuth.isSignedInWithAuthState();

authState.isAuthenticated; // true or false
authState?.accessToken; // the user's access token object
authState?.idToken; // the user's id token object

// or boolean only
const isLoggedIn = await identityAuth.isSignedIn();

// see rest of API below for more methods
```

Applications will need to be registered in Okta before they can use this library. Please contact the Identity team to do this. They will provide you with the configuration values you need to initialise the library.

In general you use the _Access Token_ to make API requests to Guardian services on behalf of the user, and you use the _ID Token_ to read the user's identity information within your application.

This is what replaces the `SC_GU_U` cookie that is used in the legacy identity platform.

## Configuration

```ts
import type { IdentityAuthOptions } from '@guardian/identity-auth';
import { IdentityAuth } from '@guardian/identity-auth';

const config: IdentityAuthOptions = {
	issuer: 'https://profile.theguardian.com/oauth2/SERVER_ID',
	clientId: 'example-client-id',
	redirectUri: 'https://theguardian.com',
	scopes: ['openid', 'profile'], // and any other scopes you need
};

const identityAuth = new IdentityAuth(config);
```

## API

`IdentityAuth` exposes the following:

- [`isSignedInWithAuthState`](#isSignedInWithAuthState)
- [`isSignedIn`](#isSignedIn)
- [`authStateManager`](#authStateManager)
  - [`getAuthState`](#getAuthState)
  - [`subscribe`](#subscribe)
  - [`unsubscribe`](#unsubscribe)
- [`tokenManager`](#tokenManager)
  - [`clear`](#clear)
  - [`getTokens`](#getTokens)
  - [`getTokensSync`](#getTokensSync)
  - [`setTokens`](#setTokens)
- [`token`](#token)
  - [`getWithoutPrompt`](#getWithoutPrompt)
  - [`verifyToken`](#verifyToken)

These features were identified during the spike for implementation with DCR. We may need to add on more features as the migration progresses.

#### `isSignedInWithAuthState`

Checks if the user is signed in, and updates the auth state as necessary, returns the current auth state

This performs side effects.

1. If the user has a GU_SO cookie, they have recently signed out, so we should clear their tokens (side effect)
2. If user tokens already exist, they are signed in
3. If the user doesn't have tokens, but they have a GU_U cookie, they are "maybe" signed in
   - We can try to get tokens without prompting/redirecting the user for credentials (side effect)
4. If the user doesn't have tokens or a GU_U cookie, they are not signed in

Returns the current auth state.

```mermaid
flowchart TD
    startNode([Start]) --> hasGuSO
    hasGuSO{GU_SO cookie\npresent?}
    hasGuSO --Yes--> clearTokens(Clear Access and\nID tokens) --> isSignedOut([User is signed out])
    hasGuSO --No--> hasTokens{Access and ID\ntokens present\nand valid?}
    hasTokens --Yes--> isSignedIn([User is signed in])
    hasTokens --No--> hasGUU{GU_U cookie\npresent?}
    hasGUU --No--> isSignedOut
    hasGUU --Yes--> isMaybeSignedIn([User is maybe signed in])
```

```ts
import { OAuthError } from '@guardian/identity-auth';

try {
	// Check if the user is logged in and return the current auth state
	const authState = await identityAuth.isSignedInWithAuthState();

	console.log(authState.isAuthenticated); // true or false
	console.log(authState?.accessToken); // the user's access token
	console.log(authState?.idToken); // the user's id token

	// e.g. read the id token claims to use in your application
	if (authState.idToken) {
		const claims = authState.idToken.claims;

		console.log('identity id', claims.legacy_identity_id);
		console.log('username', claims.identity_username);
	}

	// e.g use the access token to make an API request
	if (authState.accessToken) {
		const accessTokenString = authState.accessToken.accessToken;

		const response = await fetch('https://some-endpoint.theguardian.com', {
			headers: {
				Authorization: `Bearer ${accessTokenString}`,
			},
		});

		// Handle the response
	}
} catch (error) {
	if (error instanceof OAuthError) {
		// Handle OAuth errors
	}
	// Handle other errors
}
```

#### `isSignedIn`

Checks if the user is signed in, returns a boolean. Same as the `isSignedInWithAuthState` method, but without the auth state.

```ts
import { OAuthError } from '@guardian/identity-auth';

// Check if the user is logged in and return a boolean
try {
	const isLoggedIn = await identityAuth.isSignedIn();
} catch (error) {
	if (error instanceof OAuthError) {
		// Handle OAuth errors
	}
	// Handle other errors
}
```

### `authStateManager`

The `authStateManager` is an instance of `AuthStateManager`, which manages the auth state of the user.

The `IdentityAuthState` type is exported for convenience.

#### `getAuthState`

Returns the current auth state.

```ts
const authState = identityAuth.authStateManager.getAuthState();

authState.isAuthenticated; // true or false
authState?.accessToken; // the user's access token
authState?.idToken; // the user's id token
```

#### `subscribe`

Subscribes to changes in the auth state. The callback will be called with the new auth state whenever it changes.

```ts
import type { IdentityAuthState } from '@guardian/identity-auth';

const handler = (authState: IdentityAuthState) => {
	// do something with the new auth state
	console.log(authState);
};

identityAuth.authStateManager.subscribe(handler);
```

#### `unsubscribe`

Unsubscribes from changes in the auth state for a given handler.

```ts
identityAuth.authStateManager.unsubscribe(handler);
```

### `tokenManager`

The `tokenManager` is an instance of `TokenManager`, which manages the storage and retrieval of tokens.

#### `clear`

Clears the access and id tokens from storage.

```ts
identityAuth.tokenManager.clear();
```

#### `getTokens`

Gets the tokens from storage asynchronously, can refresh tokens if required and verify them. Returns tokens or `undefined`.

- `verifyToken` - If `true`, the tokens will be verified before being returned. Defaults to `true`.
- `refreshIfRequired` - If `true`, the tokens will be refreshed if they are expired. Defaults to `false`.

```ts
const tokens = await identityAuth.tokenManager.getTokens({
	verifyToken: true,
	refreshIfRequired: true,
});

tokens.accessToken; // the user's access token
tokens.idToken; // the user's id token
```

#### `getTokensSync`

Gets the tokens from storage synchronously, does not refresh tokens or verify them.

Returns tokens or `undefined`.

```ts
const tokens = identityAuth.tokenManager.getTokensSync();

tokens.accessToken; // the user's access token
tokens.idToken; // the user's id token
```

#### `setTokens`

Sets the tokens in local storage, very unlikely you'll need to manually use this unless you're manually retrieving tokens using the `token` class.

```ts
import type { AccessToken, IDToken } from '@guardian/identity-auth';

const accessToken: AccessToken = {
	...
}
const idToken: IDToken = {
	...
}

const tokens = {
	accessToken,
	idToken,
};

identityAuth.tokenManager.setTokens(tokens);
```

### `token`

The `token` object is an instance of `Token`, which manages the retrieval of tokens by performing the OAuth Authorization Code Flow with PKCE, exchanging the authorization code for tokens and verifying the ID and access tokens.

#### `getWithoutPrompt`

Performs the Authorization Code Flow with PKCE, exchanging the authorization code for tokens and verifying the ID token, without prompting the user and using a hidden iframe. Returns `TokenResponse` or throws an `OAuthError`.

```ts
import { OAuthError } from '@guardian/identity-auth';

try {
	const tokenResponse = await identityAuth.token.getWithoutPrompt();

	tokenResponse.tokens.accessToken; // the user's access token
	tokenResponse.tokens.idToken; // the user's id token
	tokenResponse.state; // the state parameter passed to the authorize endpoint

	// set the tokens in local storage
	identityAuth.tokenManager.setTokens(tokenResponse.tokens);
} catch (error) {
	// handle error
	if (error instanceof OAuthError) {
		// handle OAuth error
		error.error; // the error returned by the OAuth server
		error.error_description; // the error description returned by the OAuth server
		error.message; // the error message
	}
	// handle other errors
}
```

#### `verifyToken`

Verifies the ID token, checking the signature and claims, and verifies the access token using the `at_hash` claim from the IDToken.

- `idToken` - The ID token to verify
- `accessToken` - The access token to verify

Returns `void` or throws an `OAuthError`.

```ts
import { OAuthError } from '@guardian/identity-auth';

try {
	// tokens from somewhere
	const tokens = identityAuth.tokenManager.getTokensSync();

	identityAuth.token.verifyToken(tokens.idToken, tokens.accessToken);
} catch (error) {
	// handle error
	if (error instanceof OAuthError) {
		// handle OAuth error
		error.error; // the error returned by the OAuth server
		error.error_description; // the error description returned by the OAuth server
		error.message; // the error message
	}
	// handle other errors
}
```

## Types

We export a number of types for convenience, which can be useful when using the library. See the types themselves for more information.

```ts
import type {
	IdentityAuthOptions,
	IdentityAuthState,
	AccessToken,
	AccessTokenClaims,
	CustomClaims,
	IDToken,
	IDTokenClaims,
} from '@guardian/identity-auth';

// also the error which isn't a type, but is useful for checking if an error is an OAuth error
// e.g if (error instanceof OAuthError) { ... }
import { OAuthError } from '@guardian/identity-auth';
```

### Custom claims

Depending on the scopes you request, you may get custom claims in the Access Token or ID Token. If you don't create a type for these claims, then typescript will complain about them not existing.

You can add custom claims for you application by creating a type that extends the `CustomClaims` type.

A claim value can be a `string`, `number`, `boolean`, `string[]`, `number[]` or `boolean[]`.

```ts
type CustomAccessTokenClaims = CustomClaims & {
	foo: string;
};

type CustomIdTokenClaims = CustomClaims & {
	bar: number[];
	baz: boolean;
};

// pass the types to the IdentityAuth class as generics
const auth = new IdentityAuth<CustomAccessTokenClaims, CustomIdTokenClaims>({
	...
});

// the custom claims will be available on the AccessToken and IDToken types
const authState = auth.authStateManager.getAuthState();

authState.accessToken.foo; // string
authState.idToken.bar; // number[]
authState.idToken.baz; // boolean
```

## Contributing

Due to the nature of this library, if you want to make changes to it, we suggest you speak to the Identity team first regarding the nature of the changes you want to make, e.g. for bug fixes, new features, etc.


ShraddhaPande:
Need to find out the size of this bundle - will see if there is a way to find it before it actually gets published . Mahesh to reach out to CSTI team .

coldlink:
> Thanks Mahesh! I'm not sure on the details but did want to ask - would this be better as its own lib within the monorepo rather than nested in `@guardian/libs`?

I was wondering about this too. That might be a better idea, not sure how I'd go about doing that in this repo though?

I only included it in libs for now as that seemed the easiest place to develop it.

joecowton1:
@coldlink had a spare 15 mins so created an empty nx project here: https://github.com/guardian/csnx/pull/623

.github/CODEOWNERS
```diff
@@ -18,3 +18,5 @@ libs/@guardian/eslint-config/* @guardian/client-side-infra
 libs/@guardian/eslint-config-typescript/* @guardian/client-side-infra
 libs/@guardian/browserlist-config/*  @guardian/client-side-infra @guardian/dotcom-platform
 libs/@guardian/libs/* @guardian/client-side-infra @guardian/apps-rendering @guardian/dotcom-platform
+
+libs/@guardian/identity-auth/* @guardian/identity
```
coldlink, line 22: @guardian/client-side-infra possible to give @guardian/identity write access to this repo for code ownership?

georgeblahblah (approved):
This looks fantastic! I've only reviewed this from an API surface area perspective, as I don't have enough domain knowledge to assess the OAuth/Okta implementation details. I had one non-blocking comment about using a type to specify acceptable values for the event names in `emitter`.

I think using this library will establish a good boundary between DCR and identity.

We also have an open question around polyfilling `crypto.subtle`, which I believe is not implemented in Safari 10.1.

Thanks so much @coldlink!

libs/@guardian/identity-auth/package.json
```diff
@@ -1,15 +1,18 @@
 {
 	"name": "@guardian/identity-auth",
-	"version": "4.0.0",
+	"version": "0.0.1",
```
raphaelkabo, line 3: Personal preference, but I'd start versioning a library like this at `0.1.0`, since the first version defines the API rather than making minor patch changes, but the API may not be in a stable state yet, which would be `1.y.z`. Maybe CSTI has its own rules though! Cf.: https://stackoverflow.com/a/24154727

libs/@guardian/identity-auth/src/token.ts
```diff
@@ -0,0 +1,910 @@
...
+ * @returns Promise<void> - resolves when the token is verified
+ */
+export const verifyToken = async (
+	idToken: IDToken,
+	accessToken: AccessToken,
+	options: IdentityAuthOptions,
+	oauthUrls: OAuthUrls,
+) => {
+	// check if the token has already been verified
+	if (memoizedVerifyToken.has(idToken.claims.jti)) {
```
kelvin-chappell, line 602: Is there any potential here for getting around verification by marking a token as verified when it hasn't been?  I know that's a slightly vague question as I'm not sure how the memoization is working

libs/@guardian/identity-auth/src/@types/OAuth.ts
```diff
@@ -0,0 +1,128 @@
...
+	isAuthenticated: boolean;
+};
+
+/**
+ * Parameter required for the OAuth2 authorization code flow, /authorize endpoint
+ *
+ * https://developer.okta.com/docs/reference/api/oidc/#authorize
+ *
+ * In our case specifically set up for:
+ * 	- Authorization Code Flow with PKCE
```
raphaelkabo, line 41:
```suggestion
 *  - Authorization Code Flow with PKCE
```

libs/@guardian/identity-auth/src/@types/OAuth.ts
```diff
@@ -0,0 +1,128 @@
...
+ *  - Okta Post Message (iframe) response mode
```
raphaelkabo, line 42:
```suggestion
 * 	- Okta Post Message (iframe) response mode
```

libs/@guardian/identity-auth/src/@types/Token.ts
```diff
@@ -0,0 +1,175 @@
...
+ */
+export type JWTPayload<T extends CustomClaims = CustomClaims> = UserClaims &
+	T & {
+		scp?: string[];
+	};
+
+/**
+ * A JWT (JSON Web Token)
+ */
+export interface JWTObject<T extends CustomClaims = CustomClaims> {
```
raphaelkabo, line 100: Loving how neat this is.

libs/@guardian/identity-auth/src/@types/OAuth.ts
```diff
@@ -0,0 +1,128 @@
...
+/**
+ * Defines the response from the OAuth2 /token endpoint
+ *
+ * https://developer.okta.com/docs/reference/api/oidc/#response-properties-4
+ */
+export interface OAuthTokenResponse {
+	access_token: string;
+	id_token: string;
+	expires_in: number;
+	scope: string;
```
raphaelkabo, line 116: I guess there isn't a way to validate the scope response from the endpoint - I presume it's a `+`-delimited list of strings?

libs/@guardian/identity-auth/src/identityAuth.ts
```diff
@@ -0,0 +1,146 @@
...
+			if (
+				authState?.isAuthenticated &&
+				authState.accessToken &&
+				authState.idToken
+			) {
+				// validate the id token and access token to make sure auth state is still valid
+				await this.token.verifyToken(authState.idToken, authState.accessToken);
+
+				// if the id token is valid, return the auth state
+				return authState;
```
raphaelkabo, line 92: Doesn't this line always return the `authState`, irrespective of whether `this.token.verifyToken` says that the ID token is valid?

raphaelkabo: Oh, `verifyToken` throws, I see!

raphaelkabo (approved):
The OAuth implementation and the API both look great, stellar job. I left a few minor notes.

libs/@guardian/identity-auth/src/crypto.ts
```diff
@@ -0,0 +1,95 @@
...
+// The code challenge is a base64url encoded SHA256 hash of the code verifier
+export const generateCodeChallenge = async (codeVerifier: string) => {
+	// generate SHA256 hash from code verifier
+	const hash = await generateSha256Hash(codeVerifier);
+
+	// convert hash to base64url string
+	const base64url = base64UrlEncode(hash);
+
+	return base64url;
+};
```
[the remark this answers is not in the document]
coldlink: we have to use the browsers Web Crypto API, instead of the node crypto module

libs/@guardian/identity-auth/src/token.ts
coldlink: i dont think so, it's stored in a `Map`, where the only reference to it is in the method below that, so there isn't any other way of getting to that Map in memory.

libs/@guardian/identity-auth/src/crypto.ts
```diff
@@ -0,0 +1,95 @@
...
+	const utf8str = window.atob(base64);
+
+	try {
+		// decode any URI-encoded UTF-8 characters
+		return decodeURIComponent(window.escape(utf8str));
+	} catch (e) {
+		// if there was an error, just return the UTF-8 string
+		return utf8str;
+	}
+};
```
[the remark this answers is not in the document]
coldlink: Good point, I'll see if we can get away with just using `base64UrlToString`

kelvin-chappell (approved):
Great work very speedily done!  I can't see any obvious security flaws.  Thanks for going through it with us.  ✅

libs/@guardian/identity-auth/src/crypto.ts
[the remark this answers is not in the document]
coldlink: Yup, I've removed `base64UrlDecode` to save a few bytes

coldlink:
> This looks fantastic! I've only reviewed this from an API surface area perspective, as I don't have enough domain knowledge to assess the OAuth/Okta implementation details. I had one non-blocking comment about using a type to specify acceptable values for the event names in `emitter`.
> 
> I think using this library will establish a good boundary between DCR and identity.
> 
> We also have an open question around polyfilling `crypto.subtle`, which I believe is not implemented in Safari 10.1.
> 
> Thanks so much @coldlink!

@georgeblahblah here's the list of features we use and the safari version supported

[crypto.getRandomValues](https://caniuse.com/getrandomvalues) - Safari 5
[crypto.subtle.digest](https://caniuse.com/mdn-api_subtlecrypto_digest) - Safari 7
[crypto.subtle.importKey](https://caniuse.com/mdn-api_subtlecrypto_importkey) - Safari 7
[crypto.subtle.verify](https://caniuse.com/mdn-api_subtlecrypto_verify) - Safari 7
[TextEncoder](https://caniuse.com/textencoder) - Safari 10.1 (Desktop) Safari 10.3 (Mobile)

coldlink:
@joecowton1 all good to go, just need approval from CSTI now

joecowton1 (approved):

PR #597, opened 2023-05-31
Update getPackages version

joecowton1:
## What are you changing?

- updating scripts that use `@manypkg/get-packages` inline with a breaking change with requires that we pass a path into the `getPackages` function

## Why?

- we're updating the version and its a breaking change


jamesmockett (approved):
👍

PR #642, opened 2023-06-13
Update README.md as codeowners test

Georges-GNM:
## What are you changing?

- Small tweak to the readme mainly as a test of the codeowners flow

## Why?

-


georgeblahblah (approved):

joecowton1 (approved):

Georges-GNM:
> very strange that codeownership didn't work in #641
> 
> I wonder if thats because the change to the source file and change to the changeset is in the same commit in that PR?
> 
> As I noticed in this PR it's in 2 separate commits?

Hey Mahesh, sorry, missed this - my reading of this is that Identity has ownership over everything in /Identity-auth, but only CSTI have ownership over the .changeset folder, and that's what's causing the issue here. If you weren't publishing to npm identity's review would be enough, but since there is a changeset (and I suppose regardless of which commit adds it), CSTI's review is needed as well. 

Have opened a PR to amend that!

PR #651, opened 2023-06-16
add picture design type

alinaboghiu:
## What are you changing?

Adding the Picture article design type

## Why?

So that we can start to render this format in DCR

### DIFF
--- a/.changeset/wild-buttons-jog.md
+++ b/.changeset/wild-buttons-jog.md
@@ -1,0 +1,5 @@
+---
+'@guardian/libs': minor
+---
+
+Introducing a Picture design type for use by image articles
--- a/libs/@guardian/libs/src/format/ArticleDesign.ts
+++ b/libs/@guardian/libs/src/format/ArticleDesign.ts
@@ -1,5 +1,6 @@
 export enum ArticleDesign {
 	Standard,
+	Picture,
 	Gallery,
 	Audio,
 	Video,


jamesmockett:
Updating the type probably warrants a changeset and minor version bump?

alinaboghiu:
> Updating the type probably warrants a changeset and minor version bump?

Thanks, wasn't sure how this worked :D fixing

jamesmockett (approved):

mxdvl:
> > minor version bump?
> 
> Just a heads up (for the future since this has already gone out 🔮), I think this would be a **major** version bump because you're changing the existing public API?

Only because we use the magically awkward TypeScript `enum`s, though, right? If this was a `string` union I don't think it would have warranted a major bump.

JamieB-gu:
> Only because we use the magically awkward TypeScript `enum`s, though, right? If this was a `string` union I don't think it would have warranted a major bump.

I don't think it matters how you construct the union type (via `type` or `enum`); the change that we consider breaking is adding a new member to an exported type? You've widened the type and therefore the consumer's code may need to be adapted to handle the additional case.

For example:

```ts
// Library
export type Num = 'One' | 'Two';

// Consumer
import type { Num } from 'library';

const numToNumber = (num: Num): number => {
  switch (num) {
    case 'One':
      return 1;
    case 'Two':
      return 2;
  }
}
```

If the library changes the type of `Num` to:

```ts
export type Num = 'One' | 'Two' | 'Three';
```

the consumer gets a compiler error:

```
Function lacks ending return statement and return type does not include 'undefined'.
```

because `numToNumber` doesn't have a case for `'Three'` and therefore JS implicitly returns `undefined`, which is not a `number`.

This is also the case if we construct `Num` using `enum`:

```ts
enum Num {
  One,
  Two,
  Three,
}
```

PR #662, opened 2023-06-22
Remove esmodule: true from rollup config

joecowton1:
## What are you changing?

- updating rollup config

## Why?

- we introduced this in a [recent rollup upgrade](https://github.com/guardian/csnx/pull/627/files), however we're having issues with @guardian packages on DCR (which uses cjs) so this is an attempt to fix those


mxdvl (commented):
We would need to try and install these packages directly via `yarn link` in DCR before knowhing whether this’ll fix it

PR #665, opened 2023-06-22
Rolling back `npm-package` to use TS 4.9.5

joecowton1:
## What are you changing?

- rolling back `npm-package` to use TS 4.9.5.
- releasing patch versions of packages that previously contained the triple directive bug described below

## Why?

We're rolling back the version of TypeScript exclusively for the @csnx/npm-package tool, as the rollup-plugin-ts does not support TypeScript 5.0 or 5.1

This is a pragmatic decision that enables our libraries to work with the latest version of TypeScript, but not a stable position we can maintain for long…

### DIFF
--- a/.changeset/pretty-tools-end.md
+++ b/.changeset/pretty-tools-end.md
@@ -1,0 +1,6 @@
+---
+'@guardian/source-react-components': patch
+'@guardian/source-react-components-development-kitchen': patch
+---
+
+Update build output config
--- a/.changeset/pretty-tools-end.md
+++ b/.changeset/pretty-tools-end.md
@@ -1,6 +1,0 @@
----
-'@guardian/source-react-components': patch
-'@guardian/source-react-components-development-kitchen': patch
----
-
-Update build output config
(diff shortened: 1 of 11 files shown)


mxdvl:
I think that it would be better to deprecated the latest major versions and instead release this change as a patch of previous major versions, so that the changes from #627 can be incorporated.

This would results in:

- @guardian/ab-core @ `4.0.1` instead of `5.0.1`
- etc.

joecowton1:
> I think that it would be better to deprecated the latest major versions and instead release this change as a patch of previous major versions, so that the changes from #627 can be incorporated.
> 
> This would results in:
> 
> * @guardian/ab-core @ `4.0.1` instead of `5.0.1`
> * etc.

While i don't disagree this gets confusing where there have been subsequent releases ie in libs..

mxdvl (commented):
What we’ve done here is rolling back the version of TypeScript exclusively for the `@csnx/npm-package` tool, as the `rollup-plugin-ts` does not support TypeScript 5.0 or 5.1

Once the PR title and description have been updated to reflect that, happy to approve! 😁

PR #667, opened 2023-06-26
Rename `ArticlePillar` to `Pillar`

JamieB-gu:
## What are you changing?

Renamed the `ArticlePillar` type that's part of `format` to `Pillar`.

- Updated `guardian/libs` to specify and export new name
- Updated `atoms-rendering` to use new exported name
- Updated `source-react-components-development-kitchen` to use new exported name

This is a breaking change because it alters the public API of `guardian/libs`, `guardian/atoms-rendering` and `guardian/source-react-components-development-kitchen`. Consumers will have to update their code to use the new name in their imports.

## Why?

In DCR (one of the two primary consumers of `ArticleFormat`, along with AR) we support several categories of pages that aren't articles: fronts, the all newsletters page etc.. These pages still have a concept of pillar, to use in the nav and footer for example, and need to make use of this `Pillar` type. Therefore we'd like to rename it to reflect its wider usage.

For more context see the following:

- https://github.com/guardian/dotcom-rendering/issues/7562
- https://github.com/guardian/dotcom-rendering/pull/7587#discussion_r1194931855
- https://github.com/guardian/dotcom-rendering/pull/7967

## How will changesets handle this release?

Several packages in this monorepo depend on `guardian/libs` as a peer dependency. My understanding is that changesets will automatically update these and release them as major version bumps[^1] as part of the changesets release PR. If you don't think that's the case please let me know!

I'm assuming that because `atoms-rendering` and `source-react-components-development-kitchen` are already getting major bumps as part of this PR, that will be folded into the release (i.e. they won't get bumped a second time). Looking at the usages of `guardian/libs` I think these two packages are the only ones that use `format` specifically, and so should be the only ones that are actually affected by this change. Again, please let me know if you don't think that's the case.

[^1]: https://github.com/changesets/changesets/blob/main/docs/decisions.md#the-versioning-of-peer-dependencies


joecowton1:
Closed in favour of :https://github.com/guardian/csnx/pull/680

JamieB-gu:
> Closed in favour of :#680

Are these two the same? Wasn't #680 a minor change to add a new API and deprecate the old one, whereas this PR is the major change that removes the deprecated API and updates the consumers?

joecowton1:
Sorry you're right.

PR #653, opened 2023-06-20
Update youtube overlay design

abeddow91:
## What are you changing?
Updates the youtube overlay.  Updates include 
- moving the timestamp top right
- making the play button pillarless
- using the new svg play button

## Why?

This is inline with new designs that will be carried out across web (DCR, Atoms Rendering and frontend) and Apps

PR #670, opened 2023-06-27
Update youtube overlay pill to include the video categeory

sophie-macmillan:
## What are you changing?
This change builds on https://github.com/guardian/csnx/pull/653, which moves the duration pill.
If the video category is Documentary, Explainer or Live it includes this in the pill

## Why?
Its helpful to give users a bit more information about the type of video they are about to watch

| Change | Screenshot |
|--------|--------|
| Duration only (no change) | image |
| Documentary | image |
| Explainer | image |
| Live | image |
| Live, with no duration | image |

### DIFF
--- a/.changeset/bright-falcons-ring.md
+++ b/.changeset/bright-falcons-ring.md
@@ -1,0 +1,5 @@
+---
+'@guardian/atoms-rendering': minor
+---
+
+Update the YoutubeAtom's duration pill to include the videoCategory, if provided


marjisound (approved):
LGTM

libs/@guardian/atoms-rendering/src/YoutubeAtomOverlay.tsx
```diff
@@ -81,17 +84,46 @@ const playButtonStyling = css`
...
-	padding: ${space[1]}px ${space[3]}px;
 	border-radius: ${space[3]}px;
+	padding: 0 6px;
+	display: inline-flex;
 `;
 
+const pillItemStyles = css`
+	:last-child:not(:only-of-type) {
+		border-left: 1px solid rgba(255, 255, 255, 0.5);
+	}
```
mxdvl, line 102: Very clever! Might be worth adding a comment explaining what this does? Would `nth-of-type(n+2)` be more appropriate, as it targets all elements but the first.

What's `:not` support like these days?

libs/@guardian/atoms-rendering/src/YoutubeAtomOverlay.tsx
```diff
@@ -81,17 +84,46 @@ const playButtonStyling = css`
...
+			width: 9px;
+			height: 9px;
+			border-radius: 50%;
+			background-color: ${palette.news[500]};
+			display: inline-block;
+			position: relative;
+			margin-right: 0.1875rem;
+		`}
+	}
+`;
```
mxdvl, line 122: The recommended way to do emotion styling is [composing styles](https://emotion.sh/docs/@emotion/css#cx), rather than methods that take arguments.

mxdvl (approved):
🙆

libs/@guardian/atoms-rendering/src/YoutubeAtomOverlay.tsx
```diff
@@ -81,17 +84,46 @@ const playButtonStyling = css`
...
+		border-radius: 50%;
+		background-color: ${palette.news[500]};
+		display: inline-block;
+		position: relative;
+		margin-right: 0.1875rem;
+	}
+`;
+
+const capitalise = (str: string): string =>
+	str.charAt(0).toUpperCase() + str.slice(1);
```
mxdvl, line 125: If you want to narrow it down even further 😉 

```suggestion
const capitalise = <T extends string>(str: T): Capitalize<T> =>
	str.charAt(0).toUpperCase() + str.slice(1);
```

mxdvl: 👆 the main problem with this approach is that you end up having to cast the type, which feels like you’re not gaining much.

joecowton1 (approved):

arelra:
If you do come back to this it might be useful to capture this logic in a test. There is an existing `YoutubeAtom.test.tsx` (or we could spin out a `YoutubeAtomOverlay.test.tsx`) which uses react-testing-library as a lightweight way to assert on the component rendering.

PR #674, opened 2023-06-30
Identity Auth | Set up `AutoRenewService` and update type definitions for `IdentityAuthState`

coldlink:
## What does this change?

- Adds an `autoRenew` service which will automatically renew/refresh tokens silently before they expire
  - This is enabled through the `autoRenew` flag in the `IdentityAuthOptions` config object, this defaults to `true`
  - The time when it's renewed is controlled through the `renewGracePeriod` option, which defaults to `60` seconds before the tokens expire
- Updates the type definitions for `IdentityAuthState` to explicitly define the `IdentityAuthStateNotAuthenticated` and `IdentityAuthStateAuthenticated` states, which are now also exported

### `AutoRenew` service

The auto renew service is required as if a user stays on a page longer than the time it takes for a token to expire, then when a user attempts to take an authenticated action, it would behave as if the user is not logged in, e.g. posting a comment.

We can mitigate this edge case through the auto renew service, which will automatically refresh the token in the background before it expires should the user still be on the page.

We also use the Page Visibility API, so that tokens are only ever renewed on the active tab, rather than all renewing at the same time on multiple different hidden/background tabs. The refreshed tokens will migrate to the other tabs through local storage anyway.

The main reason for this is to limit the number of refresh calls that we make to Okta, without the visiblity API, if a user had 5 tabs open or in a hidden tab, then when the token would expire, it would make 5 separate API calls to Okta in order to renew.
The Page Visibility API limits this to 1 call on an open tab, or when a user comes back to a hidden/background tab.

This is difficult to automatically test, but through manual testing was done to make sure this behaviour worked as expected.

## Note

While this is a breaking change regarding the type definitions, we're making this a `minor` bump, as we don't want to release `1.0.0` until we're ready to release the migration piece on DCR.

### DIFF
--- a/libs/@guardian/identity-auth/src/autoRenew.ts
+++ b/libs/@guardian/identity-auth/src/autoRenew.ts
@@ -1,0 +1,146 @@
+import { getCookie } from '@guardian/libs';
+import type { IdentityAuthOptions, IdentityAuthState } from './@types/OAuth';
+import type { AuthStateManager } from './authState';
+import type { Emitter } from './emitter';
+
+/**
+ * @class AutoRenewService
+ * @description Managing the auto renewal of tokens, if enabled
+ */
+export class AutoRenewService {
+	#options: Required<IdentityAuthOptions>;
+	#emitter: Emitter;
+	#authStateManager: AuthStateManager;
+	started = false;
+
+	#renewAtTimeoutId: number | undefined;
+
+	constructor(
+		options: Required<IdentityAuthOptions>,
+		emitter: Emitter,
+		authStateManager: AuthStateManager,
+	) {
+		this.#options = options;
+		this.#emitter = emitter;
+		this.#authStateManager = authStateManager;
+
+		// if there is a storage event, clear the timeout, and start a new one
+		this.#emitter.on('storage', () => {
+			this.#clearRenewAtTimeout();
+			this.#handleAuthStateChange();
+		});
+	}
+
+	/**
+	 * @name clearRenewAtTimeout
+	 * @description Clears the timeout for renewing the access token, if it exists
+	 */
+	#clearRenewAtTimeout() {
+		if (this.#renewAtTimeoutId !== undefined) {
+			window.clearTimeout(this.#renewAtTimeoutId);
+			this.#renewAtTimeoutId = undefined;
+		}
+	}
+
+	/**
+	 * @name setRenewAtTimeout
+	 * @description Sets the timeout for renewing the access token based on the expiry time and the grace period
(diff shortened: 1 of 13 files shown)


coldlink:
> Are we supporting any browsers that doesn't have the Page Visibility API, and have we tested in something like Browserstack to see that the fallback behaviour is acceptable in case the API is missing? I presume without the API the user can still log in, but their tokens will never refresh?

We shouldn't need to as support is very good and lower than our recommended browser policy.

https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API#browser_compatibility

https://www.theguardian.com/help/recommended-browsers

libs/@guardian/identity-auth/src/autoRenew.ts
```diff
@@ -0,0 +1,146 @@
...
+
+		// set the timeout
+		this.#renewAtTimeoutId = window.setTimeout(() => {
+			// if the document is hidden, then we need to wait for the visibility change event before renewing the token
+			if (window.document.hidden) {
+				// if the document is hidden, then wait for the visibility change event
+				window.document.addEventListener(
+					'visibilitychange',
+					() => {
+						if (!window.document.hidden) {
```
mxdvl, line 69: in [DCR](https://github.com/guardian/dotcom-rendering/blob/59a8452ef70250e6256f55255752229027b01894/dotcom-rendering/src/components/Liveness.importable.tsx#L272) and [commercial](https://github.com/guardian/commercial/blob/e556b3273771a0b5d672792d7e55d348282f86c3/src/core/send-commercial-metrics.ts#L158) we use [`document.visibilityState`](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState)

coldlink: updated to use `document.visibilityState` which is easier to read!

raphaelkabo (approved):
From an Identity perspective, these changes look really solid and carefully thought-through. It's a shame we can't auto-integration-test it reliably, since it might be a difficult issue to debug if we have users experiencing problems with the functionality down the line and it would be good to rule out anything testable. But until we can make that work, this looks great!

Georges-GNM (approved):
LGTM!

georgeblahblah:
> This looks pretty thorough and well tested.
> 
> Wondering that if there are aspects of this functionality that could be covered by a [library like SWR](https://swr.vercel.app/) if we took (P)react as a given?

@mxdvl I think we'll consume this library in frontend in areas where we don't always (or ever?) have Preact

georgeblahblah (approved):

PR #675, opened 2023-06-30
feat(YoutubeAtom): player requires ad targeting

mxdvl:
## What are you changing?

- The `YoutubeAtomPlayer` can only be initialised once, and we require the ad targeting to be present before initialisation.

## Why?

- Ad targetting is set asyncronously client-side in https://github.com/guardian/dotcom-rendering/pull/8095, so we need to handle the value being `undefined` for the first render.
  - The current interface wasn’t srict enough so thanks @arelra for spotting this as the compiler would not have complained!


arelra (approved):
lgtm!

ob6160 (approved):

PR #682, opened 2023-07-05
Fix ad targeting for YoutubeAtom stories 

arelra:
## What are you changing?

Pass valid ad targeting to `YoutubeAtom` stories so the player is able to load.

## Background

Since https://github.com/guardian/dotcom-rendering/pull/8095 ad targeting will be set async on the client so it could be undefined initially and then defined in subsequent render passes.

However the player requires ad targeting before it initialises and makes the request to YouTube so we block the player until ad targeting is defined as implemented in #675.

### DIFF
--- a/libs/@guardian/atoms-rendering/src/YoutubeAtom.stories.tsx
+++ b/libs/@guardian/atoms-rendering/src/YoutubeAtom.stories.tsx
@@ -2,6 +2,7 @@ import { ArticlePillar } from '@guardian/libs';
 import { useState } from 'react';
 import { consentStateCanTarget } from './fixtures/consentStateCanTarget';
 import { YoutubeAtom } from './YoutubeAtom';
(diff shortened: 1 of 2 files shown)


libs/@guardian/atoms-rendering/src/YoutubeAtom.stories.tsx
```diff
@@ -2,6 +2,7 @@ import { ArticlePillar } from '@guardian/libs';
 import { useState } from 'react';
 import { consentStateCanTarget } from './fixtures/consentStateCanTarget';
 import { YoutubeAtom } from './YoutubeAtom';
+import { AdTargeting } from 'dist/libs/@guardian/atoms-rendering/cjs/types-a8b0938f';
```
JamieB-gu, line 5: Should this be?

```suggestion
import type { AdTargeting } from './types';
```

JamieB-gu (approved):

Jakeii (approved):

PR #683, opened 2023-07-05
Use NX Cloud

joecowton1:
## What are you changing?

- reintroducing NX Cloud as per recipe here: https://nx.dev/recipes/ci/monorepo-ci-github-actions
- the token in `nx.json` is read-only as recommended by NX: [nx.dev/nx-cloud/account/scenarios#security-scenarios](https://nx.dev/nx-cloud/account/scenarios#security-scenarios)

## Why?

- cloud caching improves both CI times and local dev experience

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,9 @@ on:
     types: [opened, labeled, synchronize]
 
 env:
+  # The `accessToken` in `nx.json` is a read-only token for use in dev.
+  # This overrides it with a read-write token so that CI run results are 
+  # remotely cached.
   NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
 
 jobs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,9 @@ on:
   pull_request:
     types: [opened, labeled, synchronize]
 
+env:
+  NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
+
 jobs:
   validate:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


nx.json
```diff
@@ -18,17 +18,17 @@
...
 					"build",
 					"lint",
 					"test",
 					"e2e",
 					"build-storybook"
-				]
-			},
-			"runner": "nx-cloud"
+				],
+				"accessToken": "ZDI5NGQyNWMtYzRiNS00YjM0LWJkNDItMWEzMGQ2NWMwNWJkfHJlYWQtb25seQ=="
```
joecowton1, line 30: This token is read-only as recommended by NX: https://nx.dev/nx-cloud/account/scenarios#security-scenarios

.github/workflows/ci.yml
```diff
@@ -19,6 +22,8 @@ jobs:
           - task: build-storybooks
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
OllysCoding, line 26: What does this do?

OllysCoding (approved):
LGTM :)

.github/workflows/ci.yml
jamesmockett: By default the checkout action only fetches a single commit. `fetch-depth: 0` gets the full Git history. The Chromatic action needs this to keep track of changes so I assume it's the same for nx caching.

joecowton1: yeah thats my understanding

.github/workflows/ci.yml
```diff
@@ -6,6 +6,9 @@ on:
   pull_request:
     types: [opened, labeled, synchronize]
 
+env:
+  NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
```
sndrs, line 10: just adding an explanation of what's going on here (i had to have a quick think)

```suggestion
  # The `accessToken` in `nx.json` is a read-only token for use in dev.
  # This overrides it with a read-write token so that CI run results are 
  # remotely cached.
  NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
```

sndrs (approved):
great!

PR #687, opened 2023-07-07
Enforce healthy `package.json`

mxdvl:
## What are you changing?

- Enforce healthy `package.json`

## Why?

- There are [great recommendation on NPM packages](https://github.com/guardian/recommendations/blob/main/npm-packages.md), and [we now have a tool](https://github.com/guardian/actions-npm-dependencies/) that can ensure they are followed.

### DIFF
--- a/tools/scripts/deno/package-health.ts
+++ b/tools/scripts/deno/package-health.ts
@@ -1,4 +1,4 @@
-import { package_health } from 'https://deno.land/x/package_health@0.5.0/src/main.ts';
+import { package_health } from 'https://deno.land/x/package_health@0.5.1/src/main.ts';
 import { walk } from 'https://deno.land/std@0.193.0/fs/mod.ts';
 import {
 	dirname,
@@ -27,4 +27,5 @@ for await (const { path } of walk(root, {
 
 if (errors > 0) {
 	console.info(`There are ${errors} issues with \`package.json\` files…`);
+	Deno.exit(1);
 }
--- a/tools/scripts/deno/package-health.ts
+++ b/tools/scripts/deno/package-health.ts
@@ -1,0 +1,30 @@
+import { package_health } from 'https://deno.land/x/package_health@0.5.0/src/main.ts';
+import { walk } from 'https://deno.land/std@0.193.0/fs/mod.ts';
+import {
+	dirname,
+	fromFileUrl,
+	globToRegExp,
+	resolve,
+} from 'https://deno.land/std@0.193.0/path/mod.ts';
+
+const root = resolve(dirname(fromFileUrl(import.meta.url)), '..', '..', '..');
+
+let errors = 0;
+
+for await (const { path } of walk(root, {
+	includeDirs: false,
+	match: [
+		globToRegExp(root + '/package.json'),
+		globToRegExp(root + '/apps/**/package.json', { globstar: true }),
+		globToRegExp(root + '/libs/**/package.json', { globstar: true }),
+	],
+})) {
+	const contents = JSON.parse(await Deno.readTextFile(path));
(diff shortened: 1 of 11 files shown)


tools/scripts/deno/package-health.ts
```diff
@@ -0,0 +1,31 @@
+import { package_health } from 'https://deno.land/x/package_health@0.5.1/src/main.ts';
```
sndrs, line 1: what is this?

mxdvl: https://github.com/guardian/actions-npm-dependencies

It's this — might be worth turning into a proper action before consumption?

sndrs: ah ha got you 👍 

going to reply on the main thread so the discussion isn't lost when this is resovled...

sndrs (changes requested):
_these comments are largely in response to https://github.com/guardian/csnx/pull/687#pullrequestreview-1521900305, but there's probably value in having a general conversation here rather than attached to a line of code, hence moving it._

@mxdvl said:
> `https://deno.land/x/package_health@0.5.1/src/main.ts` is the package linter I've been working on — might be worth turning into a proper action before consumption? Is it acceptable to have a dependency on Deno for a make? Should it be cross-published to NPM via `dnt`?

there's a few things here:

1. it wasn't clear this was using something we owned. we should probably have think about how we publish guardian stuff for deno. is there a process for name-spacing there or is it all under `deno.land/x/`? we could always run the script with deno from the repo directly, making the link to one of our repos explicit.

2. it's not clear what it's checking. it could be useful to expose settings (maybe everything should be opt in?), because it's very hard to tell from this specific change exactly what is going to be enforced

3. regarding running it from the `makefile` vs action etc. yeah i agree if it's published as an action, i think it would make most sense to consume it that way in a project (rather than reusing the script from an action in another action). the issue then is that it bites you only once you've pushed. so then making it part of a general validation step (part of [`make lint`](https://github.com/guardian/csnx/blob/ba8751e33ac468c9605f207124b2051b7eefb0ef/Makefile#L32-L37)?) makes sense.

4. making deno a dependency: i'm actually personally quite in favour of switching the repo's scripts to deno, and of preferring that over Nx plugins. but that's def not how it works right now. that feels like a change that should made in a separate PR that is about using deno, rather than as a side effect of this change

given all that, if validating the `package.json`s is important, i think we should make sure it happens as soon as possible, and with at little friction as possible (see [the node version check](https://github.com/guardian/csnx/blob/ba8751e33ac468c9605f207124b2051b7eefb0ef/tools/scripts/check-node-version) for an example). ideally this stuff is invisible until you need to know about it.  

the possibility that i _could_ have checked before pushing is not the same as knowing i could, or that maybe i should.

same goes for needing deno - it's fine if that's an explicit condition of working in the repo, and in turn that makes sense if its widely used. 

right now, a very possible user journey would be:

- update a package
- push it
- build fails with message about invalid package
- discover `make package-health`
- fix it, run `make package-health`
- get `bash: deno: command not found` error
- ?

so in terms of this PR itself, maybe you could add the linter as a repo script, similar to [`check-packages-for-tslib`](https://github.com/guardian/csnx/blob/ba8751e33ac468c9605f207124b2051b7eefb0ef/tools/scripts/check-packages-for-tslib.mjs), or maybe better as a lib that lives in csnx and is published to NPM like the eslint configs etc (the idea of csnx is to centralise this stuff, after all).

that is then also usable in deno (though i guess there's less reason to do so).

i think there's a larger question around use of deno (possibly including writing packages as deno libs that we can then publish to npm with `dnt`), but it feels to me like it it should be a deliberate change, rather than something that sneaks in

PR #692, opened 2023-07-12
Add 'key' and 'length' methods to Storage

bryophyta:
## What are you changing?

Add wrappers for the built-in `key` method and `length` property on `Storage` objects.

## Why?

These can be useful for extracting the keys from storage, especially given that we (presumably by design?) don't provide direct access to the whole storage object via this API.

Adding `key` and `length` to the API allows users of the library to use these methods with the same extra checks that they get for other methods that wrap built-in functionality in this lib.

In light of the discussion below, I thought it was best to revise this PR to add wrappers around these built-in methods, rather than adding a new abstraction. I thought it best to add these changes on top of the original branch, but happy to rewrite the history to remove the first two commits (which add code which is then removed in the third commit) if desired.

## Original PR description

```
## What are you changing?

Adds a method to return all keys of the storage object in `lib/storage`, by iterating over them.

## Why?

This is not a built-in method on the Storage objects, but provides an abstraction over the built-in `key` method. The `key` method doesn't have that much use outside of getting all the keys, and it has a potential gotcha insofar as the order of keys isn't reliable across user-agents (but `.key()` gets keys by index).

If we aren't happy with adding this new method, could we perhaps consider giving access to the built-in `key` method on the Storage class in this library, to allow users of the library to create their own `keys`-style implementation on top of it?


```

### DIFF
--- a/libs/@guardian/libs/src/storage/README.md
+++ b/libs/@guardian/libs/src/storage/README.md
@@ -14,19 +14,19 @@ import { storage } from '@guardian/libs';
 
 Has a few advantages over the native API:
 
--   fails gracefully if storage is not available
--   you can save and retrieve any JSONable data
--   stored items can expire
+- fails gracefully if storage is not available
+- you can save and retrieve any JSONable data
+- stored items can expire
 
 _n.b. the examples below use `storage.local`, but all methods are available for both `storage.local` and `storage.session`._
 
 ## Methods
 
--   [`get(key)`](#getkey)
--   [`set(key, value, expires?)`](#setkey-value-expires)
--   [`remove(key)`](#removekey)
--   [`clear()`](#clear)
--   [`isAvailable()`](#isavailable)
+- [`get(key)`](#getkey)
+- [`set(key, value, expires?)`](#setkey-value-expires)
+- [`remove(key)`](#removekey)
+- [`clear()`](#clear)
+- [`isAvailable()`](#isavailable)
 
 ## `get(key)`
 
@@ -74,18 +74,18 @@ Optional expiry date for this item.
 
 ```js
 storage.local.set('my-item', {
-    prop1: 'abc',
-    prop2: 123,
+	prop1: 'abc',
+	prop2: 123,
 });
 
 storage.local.set(
-    'my-expiring-item',
-    {
-        prop1: 'abc',
-        prop2: 123,
-    },
-    // expires 24 hours from now
-    Date.now() + 60 * 60 * 24 * 1000,
+	'my-expiring-item',
+	{
+		prop1: 'abc',
+		prop2: 123,
+	},
+	// expires 24 hours from now
+	Date.now() + 60 * 60 * 24 * 1000,
 );
 ```
 
@@ -117,6 +117,32 @@ Removes all items from storage.
 
 ```js
 storage.local.clear();
+```
+
+## `key(index: number)`
+
+Returns: `string[] | null`
+
+Get the name of the key at `index` in the storage object. Returns `null` if index
+is out of range or storage is unavailable.
+
+### Example
+
+```js
+storage.local.key(1);
+```
+
+## `length()`
+
+Returns `number | null`
+
+Get the number of items in the storage object. Returns `null` if storage is
+unavailable.
+
+### Example
+
+```js
+storage.local.length();
 ```
 
 ## `isAvailable()`
--- a/libs/@guardian/libs/src/storage/storage.ts
+++ b/libs/@guardian/libs/src/storage/storage.ts
@@ -98,6 +98,32 @@ class StorageFactory {
 	setRaw(key: string, value: string): void {
 		return this.#storage?.setItem(key, value);
 	}
+
+	/**
+	 * Get key by index.
+	 * @returns the name of the key at that index (`string`), or `null` (if
+	 *      index is out of range of if storage is unavailable) - if the index
+	 *      is out of range, or if storage is not available.
+	 *
+	 * Wrapper for the `key` method on the native `Storage` object, with
+	 * additional check for availability of storage. Note that the _order_ of
+	 * keys in storage is dependent on user-agent implementation, and so you
+	 * should not assume that keys will be stored in any particular order (e.g.
+	 * order of insertion).
+	 * @see https://developer.mozilla.org/en-US/docs/Web/API/Storage/key
+	 */
+	key(index: number): string | null {
+		return this.#storage?.key(index) ?? null;
+	}
+
+	/**
+	 * Get the number of items in storage.
+	 * @returns the number of items in storage, or `null` if storage is
+	 * 	not available.
+	 */
+	length(): number | null {
+		return this.#storage?.length ?? null;
+	}
 }
 
 /**
--- a/libs/@guardian/libs/src/storage/storage.test.ts
+++ b/libs/@guardian/libs/src/storage/storage.test.ts
@@ -131,4 +131,44 @@ describe.each([
 		implementation.setRaw('raw item', '🦁');
 		expect(native.getItem('raw item')).toBe('🦁');
 	});
+
+	it(`gets keys in storage by index`, async () => {
+		expect(implementation.key(0)).toEqual(null);
+
+		native.setItem('item 1', 'some val');
+		native.setItem('item 2', 'some other val');
+		/**
+		 * The underlying native API doesn't guarantee order across different environments,
+		 * so we can't know for sure which key we will retrieve.
+		 */
+		expect(['item 1', 'item 2']).toContain(implementation.key(0));
+
+		native.removeItem('item 1');
+		expect(implementation.key(0)).toEqual('item 2');
+
+		setSpy.mockImplementation(functionThatThrowsAnError);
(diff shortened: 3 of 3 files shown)


libs/@guardian/libs/src/storage/storage.ts
```diff
@@ -98,6 +98,38 @@ class StorageFactory {
...
+	 * Get all of the keys in storage.
+	 *
+	 * @returns an array of all keys in the storage object,
+	 * or undefined if storage is not available.
+	 *
+	 * @note This is not a method in the native API, and gets
+	 * all keys by iterating over the entire storage object.
+	 * It is therefore potentially subject to race conditions
+	 * and if other code is modifying storage at the same time
+	 * then the results may be inaccurate.
```
mxdvl, line 112: @tjmw reminded me that JavaScript ensure run-to-completion so if the methods are synchronous it's unlikely that such a race condition may occur.

bryophyta: This is a good point! Although because the storage objects are browser features I wasn't sure if we could guarantee that the storage side of the operation would be implemented synchronously..

libs/@guardian/libs/src/storage/storage.ts
```diff
@@ -98,6 +98,38 @@ class StorageFactory {
...
+			return undefined;
+		}
+		const allKeys: string[] = [];
+		for (let i = 0; i < this.#storage.length; i++) {
+			const maybeKey = this.#storage.key(i);
+			if (maybeKey) {
+				allKeys.push(maybeKey);
+			}
+		}
+		return allKeys;
```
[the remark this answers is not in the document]
bryophyta: Unfortunately storage isn't iterable. As far as I'm away the `key` method is the only supported method for getting keys out of it?

libs/@guardian/libs/src/storage/storage.ts
```diff
@@ -98,6 +98,38 @@ class StorageFactory {
...
+	 * and if other code is modifying storage at the same time
+	 * then the results may be inaccurate.
+	 *
+	 * @note This method depends on the `key` method on the native API,
+	 * which gets keys by index. But the _order_ of keys in storage is
+	 * dependent on user-agent implementation, and so you should not rely
+	 * on the order of keys returned by this method. (See
+	 * https://developer.mozilla.org/en-US/docs/Web/API/Storage/key for more.)
+	 */
+	keys(): string[] | undefined {
```
sndrs, line 120: couple of thoughts:

1. how do you feel about something more explicit, like `allKeys()`? 
2. should we also add `key()` to our api? it only exists to discourage footguns of using the native one. forcing people to use the native one (if they need `key()`) seems wrong

libs/@guardian/libs/src/storage/README.md
```diff
@@ -119,6 +119,31 @@ Removes all items from storage.
...
 
+## `keys()`
+
+Returns: `string[] | undefined`
+
+Get all of the keys in storage.
+
+### Notes
+
+This is not a method in the native API, and gets all keys by iterating over the
```
sndrs, line 130: i think it's ok not to mention this. we have other methods that aren't (although the idea of matching the API is intriguing... we just don't already)

libs/@guardian/libs/src/storage/README.md
```diff
@@ -119,6 +119,31 @@ Removes all items from storage.
...
+entire storage object. It is therefore potentially subject to race conditions
```
sndrs, line 131: > It is therefore potentially subject to race conditions

this makes me wonder if it's something we should have. is it ok to offer a method that might return false information? how is a user supposed to know that? especially if your reading an existing usage that someone else (who did know) wrote?

is it better to some how force ourselves to confront that at the point of doing it, rather than wrapping it in something that _appears_ trustworthy (but can't be)?

libs/@guardian/libs/src/storage/README.md
```diff
@@ -119,6 +119,31 @@ Removes all items from storage.
...
+and if other code is modifying storage at the same time then the results may be
+inaccurate.
+
+This method depends on the `key` method on the native API, which gets keys by
+index. But the _order_ of keys in storage is dependent on user-agent
+implementation, and so you should not rely on the order of keys returned by this
+method. (See https://developer.mozilla.org/en-US/docs/Web/API/Storage/key for
+more.)
```
sndrs, line 139:
```suggestion
```

Given this describes the implementation, rather than the feature, I'd be inclined to leave this information out of the API description. 

One of the benefits of an abstracted API like this is that you can use it without worrying about what's happening below the surface. You only really need to know that stuff when working on (rather than with) the API itself.

putting this info at the point at which it's used could be very handy, but at this point is maybe distracting rather than informative?

libs/@guardian/libs/src/storage/storage.ts
```diff
@@ -98,6 +98,38 @@ class StorageFactory {
...
+	 * all keys by iterating over the entire storage object.
+	 * It is therefore potentially subject to race conditions
+	 * and if other code is modifying storage at the same time
+	 * then the results may be inaccurate.
+	 *
+	 * @note This method depends on the `key` method on the native API,
+	 * which gets keys by index. But the _order_ of keys in storage is
+	 * dependent on user-agent implementation, and so you should not rely
+	 * on the order of keys returned by this method. (See
+	 * https://developer.mozilla.org/en-US/docs/Web/API/Storage/key for more.)
```
sndrs, line 118:
```suggestion
```

as with my comment above, it might be useful to move this info to the part of the code where `key` is used. it's certainly useful for people working on the API in the future, but doesn't particularly help people who are using our method (what can they actually do with that info?)

libs/@guardian/libs/src/storage/storage.ts
```diff
@@ -98,6 +98,38 @@ class StorageFactory {
...
+	 */
+	keys(): string[] | undefined {
+		if (!this.#storage) {
+			return undefined;
```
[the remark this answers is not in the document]
bryophyta: I did wonder about this.. I was thinking that it was worth signalling the difference between there not being any keys in the store, and the store not being available. But would be happy to refactor and return `string[]` in all cases if people feel it's worthwhile 👍

libs/@guardian/libs/src/storage/README.md
bryophyta: I'll investigate a little more to find out how realistic this issue is.

If we are concerned about this then that would speak in favour of replacing this method with a simple wrapper around the built-in `key` method, as discussed above 👍

libs/@guardian/libs/src/storage/storage.ts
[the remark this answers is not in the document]
mxdvl: Just for your information, if you want to avoid mutation you can create an array of a specific length like so:

```ts
return Array.from({ length: this.#storage.length }, (_, i) =>
    this.#storage?.key(i),
    ).filter(isNonNullable);
```

libs/@guardian/libs/src/storage/README.md
bryophyta: (partial update here: https://github.com/guardian/csnx/pull/692#discussion_r1261365972)

bryophyta:
@sndrs I've rewritten this to just add wrappers around `key` and `length`, because it seemed from the above that this would be more straightforward than adding a new abstraction. What do you think?

libs/@guardian/libs/src/storage/storage.ts
```diff
@@ -98,6 +98,32 @@ class StorageFactory {
...
+	 * order of insertion).
+	 * @see https://developer.mozilla.org/en-US/docs/Web/API/Storage/key
+	 */
+	key(index: number): string | null {
+		return this.#storage?.key(index) ?? null;
+	}
+
+	/**
+	 * Get the number of items in storage.
+	 * @returns the number of items in storage, or `undefined` if storage is
```
bryophyta, line 121: I've had the `key` method return `null` if storage is unavailable, because the native method can return `null`, and it seemed best not to introduce a third possible return type. But it generally feels more semantic to return `undefined` if the storage object isn't available so I've used `undefined` for that in the `length` method.

jamesmockett:
Hi @sndrs and @mxdvl 👋  As you've both made suggestions on this PR are you happy if we approve it in its current form?

libs/@guardian/libs/src/storage/storage.ts
mxdvl: I would actually return `null` or `undefined` everywhere. While I favour `undefined` in general, I think using `null` here would be more consistent.

mxdvl (approved):
I think this is a worthwhile addition, but would favour a consistent API that returns `null` for all methods when storage is unavailable.

jamesmockett (approved):
I'll leave it to Max and you to decide whether you want to change the API, but happy to add an approval from CSTI 👍

libs/@guardian/libs/src/storage/storage.ts
bryophyta: okay thanks, happy to make this change!

bryophyta:
thanks @mxdvl and @jamesmockett ! I've changed the `length` method to return `null` as Max suggested, and tidied up the commit history 👍

PR #695, opened 2023-07-13
Update packages with dependency on Semver

joecowton1:
## What are you changing?

- updates all packages which rely on Semver to latest versions
- add pnpm override to ensure we're using safe versions of Semver in transitive deps, these can be removed once packages are all updated. 

## Why?

- security


libs/@guardian/atoms-rendering/package.json
```diff
@@ -33,9 +33,9 @@
 	},
 	"peerDependencies": {
 		"@emotion/react": "^11.11.1",
-		"@guardian/commercial": "^10.0.0",
-		"@guardian/consent-management-platform": "^13.5.0",
-		"@guardian/libs": "^15.1.0",
+		"@guardian/commercial": "^10.5.0",
+		"@guardian/consent-management-platform": "^13.6.0",
+		"@guardian/libs": "^15.2.0",
```
mxdvl, line 38: I believe that technically, any change in the peer dependencies would result in a major bump in the consuming library…

mxdvl (commented):
FYI

libs/@guardian/atoms-rendering/package.json
joecowton1: Good point, rather than confuse things lets bump peer deps later.

PR #696, opened 2023-07-14
Remove `axios` in favour of native `fetch`

jamesmockett:
## What are you changing?

- Removes `axios` from the `source-react-components` package and replaces usage with native `fetch`.

## Why?

- As we're now using Node 18 which supports native `fetch` there's no need to rely upon an external dependency for data fetching, particularly as our needs are fairly minimal.
- Native `fetch` is also faster: execution time for the `create-icons` script has been reduced from 90 seconds to 26 seconds on average.

```bash
time pnpm --filter @guardian/source-react-components create-icons
```

| Run | `axios` | `fetch` |
| --- | --- | --- |
| 1 | 1m32.532s |   0m27.927s |
| 2 | 1m31.135s | 0m26.212s |
| 3 | 1m29.709s | 0m25.295s |

### DIFF
--- a/libs/@guardian/source-react-components/scripts/create-icons/get-svgs-from-figma.ts
+++ b/libs/@guardian/source-react-components/scripts/create-icons/get-svgs-from-figma.ts
@@ -1,4 +1,3 @@
-import axios from 'axios';
 import 'dotenv/config';
 
 interface FigmaComponentsResponse {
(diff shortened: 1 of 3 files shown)


libs/@guardian/source-react-components/scripts/create-icons/get-svgs-from-figma.ts
```diff
@@ -31,15 +30,21 @@ const FIGMA_API_OPTIONS = {
 	},
 };
 
+const figmaApi = async <T>(url: string): Promise<T> => {
+	const response = await fetch(
+		`https://api.figma.com/v1/${url}`,
+		FIGMA_API_OPTIONS,
+	);
+	const data = await response.json();
+	return data;
```
mxdvl, line 39: I think this would achieve the same thing, as promises are always unwrapped:

```suggestion
	return response.json();
```

mxdvl (approved):
One less dependency! 🎉

joecowton1 (approved):

PR #703, opened 2023-07-17
Identity Auth | Change token storage/verification, and refactor certain methods

coldlink:
## What are you changing?

- We now only store the raw access token string under `gu.access_token`, and the raw id token string + nonce user
`gu.id_token` rather than the entire decoded token.
  - This is because it was possible to modify what was in local storage (e.g. a token's claims), verify the tokens, and use the modified values in your application, which can be a security issue.
  - By only storing the minimum information in local storage we can eliminate this risk, and instead
generate what is required when the tokens are decoded from local storage.
  - To facilitate this we added a public `decodeTokens` method to the `Token` class.
- We also rename `verifyToken` to `verifyTokens` as that better describes what is happening, i.e verifying both the access and id tokens.
- We also use access token timestamps to validate expiry instead of id tokens
  - This is because the ID token in Okta is only valid for 1 hour with no way to change this, whereas the access token can be valid anywhere from 5 minutes to 24 hours, so we use the access token as the source of truth for the expiry timestamp

This is technically a breaking change, but since we're still in pre-release (i.e v0) this will bump the minor version rather than releasing a v1.

### DIFF
--- a/.changeset/olive-seals-give.md
+++ b/.changeset/olive-seals-give.md
@@ -1,0 +1,8 @@
+---
+'@guardian/identity-auth': minor
+---
+
+- Add `decodeTokens` method to `Token` class
+- Refactor library so that we only store required values in l `verifyTokens`ocal storage, and now derive other values from decoding the token
+- Rename `verifyToken` method to `verifyTokens` in `Token` class to better reflect functionality
+- Use access token to verify timestamps, instead of id tokens
--- a/.changeset/olive-seals-give.md
+++ b/.changeset/olive-seals-give.md
@@ -3,6 +3,6 @@
 ---
 
 - Add `decodeTokens` method to `Token` class
-- Refactor library so that we only store required values in l `verifyTokens`ocal storage, and now derive other values from decoding the token
+- Refactor library so that we only store required values in `verifyTokens` local storage, and now derive other values from decoding the token
 - Rename `verifyToken` method to `verifyTokens` in `Token` class to better reflect functionality
 - Use access token to verify timestamps, instead of id tokens
--- a/libs/@guardian/identity-auth/src/token.ts
+++ b/libs/@guardian/identity-auth/src/token.ts
@@ -23,6 +23,7 @@ import type {
 	JWTObject,
 	JWTPayload,
 	TokenResponse,
+	Tokens,
 } from './@types/Token';
 import {
 	base64UrlEncode,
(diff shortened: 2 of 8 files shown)


.changeset/olive-seals-give.md
```diff
@@ -0,0 +1,8 @@
+---
+'@guardian/identity-auth': minor
+---
+
+- Add `decodeTokens` method to `Token` class
+- Refactor library so that we only store required values in l `verifyTokens`ocal storage, and now derive other values from decoding the token
```
kelvin-chappell, line 6:
```suggestion
- Refactor library so that we only store required values in `verifyTokens` local storage, and now derive other values from decoding the token
```

libs/@guardian/identity-auth/src/tokenManager.ts
```diff
@@ -133,20 +161,20 @@ export class TokenManager<
 	 * @name getTokens
 	 * @description Gets the tokens from storage asynchronously, can refresh tokens if required and verify them
 	 *
-	 * @param verifyToken - If true, will verify the token before returning (default: true)
+	 * @param verifyTokens - If true, will verify the token before returning (default: true)
```
kelvin-chappell, line 164:
```suggestion
	 * @param verifyTokens - If true, will verify the tokens before returning (default: true)
```

libs/@guardian/identity-auth/src/token.ts
```diff
@@ -224,10 +225,138 @@ const decodeToken = <T extends CustomClaims = CustomClaims>(
...
+			error_description: 'Invalid Access token',
+			message: 'Clock skew too large',
+		});
+	}
+
+	// construct the access token
+	const accessToken: AccessToken<AC> = {
+		accessToken: accessTokenRaw,
+		claims: accessTokenPayload,
+		expiresAt: accessTokenPayload.exp - accessTokenPayload.iat + now, // adjusting expiresAt to be in local (machine) time, to account for clock skew, // adjusted to local (machine) time, to account for clock skew
```
kelvin-chappell, line 286:
```suggestion
		expiresAt: accessTokenPayload.exp - accessTokenPayload.iat + now, // adjusting expiresAt to be in local (machine) time, to account for clock skew
```

kelvin-chappell (approved):
Spotted some typos but change looks good from my point of view 👍

kelvin-chappell (approved):

PR #721, opened 2023-07-18
Extend youtube atom API for autopausing and text overlay

abeddow91:
## What are you changing?
Extends the YoutubeAtom API to include a showTextOverlay prop which renders the headline and kicker on a faded opacity overlay.  It also extends the API to allow autopausing when a playing video is swiped offscreen. 

## Why?
This is to help build the video container on DCR fronts.

As the youtube atom currently manages overlay hiding on click, this seemed like the most logical place to keep this text overlay too, so we can use the same functionality. There is an argument to say that we shouldn't couple container and atom too closely, however, it is reasonable that this overlay might want to be used outside of the video container. Also, there are plans to move this atom into the DCR repo (as it is only used in DCR) as which point the coupling becomes a little looser still. 

## Screenshots
Screenshot 2023-07-27 at 09 47 59

Example of the autopause is available in storybook

### DIFF
--- a/libs/@guardian/atoms-rendering/src/YoutubeAtom.stories.tsx
+++ b/libs/@guardian/atoms-rendering/src/YoutubeAtom.stories.tsx
@@ -290,6 +290,7 @@ export const Sticky = (): JSX.Element => {
 				imaEnabled={false}
 				abTestParticipations={{}}
 				adTargeting={adTargeting}
+				shouldPauseOutOfView={true}
 			/>
 			<div style={{ height: '1000px' }}></div>
 		</div>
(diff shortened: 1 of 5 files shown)


libs/@guardian/atoms-rendering/src/YoutubeAtom.stories.tsx
```diff
@@ -1,4 +1,4 @@
-import { ArticlePillar } from '@guardian/libs';
```
abeddow91, line 1: ArticlePillar is deprecated so I've replaced with Pillar as per the recommendation

libs/@guardian/atoms-rendering/src/YoutubeAtom.tsx
```diff
@@ -58,6 +61,10 @@ export const YoutubeAtom = ({
 	imaEnabled,
 	abTestParticipations,
 	videoCategory,
+	kicker,
+	pillar,
+	shouldPauseOutOfView,
```
arelra, line 66: If this is undefined should we default to false here? Then it can become a required prop to `YoutubeAtomSticky.tsx`

abeddow91: good point, I'll update this 👍

PR #729, opened 2023-07-20
Helper to measure durations and create `PerformanceMeasure`s

mxdvl:
## What are you changing?

- Add a new method that can be user to measure durations.
- Using a curried function ensures the measurment can only be stopped afer it was started.

## Why?

- We rely on measurements for analysing performance and user metrics, so creating a single API can help teams with consistency.
- `PerformanceMeasure` contains all the information that `PerformanceMark` have, with an added duration. This API prevents potential bugs where two marks share the same name.

### DIFF
--- a/.changeset/spicy-pugs-smell.md
+++ b/.changeset/spicy-pugs-smell.md
@@ -1,0 +1,6 @@
+---
+'@guardian/libs': minor
+---
+
+Add the `startPerformanceMeasure` helper, which has a unified API for measuring
+durations and producing `PerformanceMeasure`s.
--- a/libs/@guardian/libs/src/performance/startPerformanceMeasure.ts
+++ b/libs/@guardian/libs/src/performance/startPerformanceMeasure.ts
@@ -1,0 +1,46 @@
+import type { TeamName } from '../logger/@types/logger';
+import type {} from './@types/measure';
+import { serialise } from './serialise';
+
+/** For browser which do not fully support the performance API */
+const fallback: ReturnType<typeof startPerformanceMeasure> = {
+	endPerformanceMeasure: () => -1,
+};
+
+/**
+ * Helper to measure the duration between two events.
+ *
+ * Once ended, measures are appended to the list of `PerformanceMeasure`
+ * and can be retrieved with `performance.getEntriesByType('measure')`
+ *
+ * @example
+ * const { endPerformanceMeasure } = startPerformanceMeasure('dotcom', 'fetch');
+ * await fetch('https://www.theguardian.com/uk.json');
+ * const duration = endPerformanceMeasure();
+ */
+export const startPerformanceMeasure = (
+	team: TeamName,
+	name: string,
+	action?: string,
+): { endPerformanceMeasure: () => number } => {
+	if (!('getEntriesByName' in window.performance)) {
+		return fallback;
+	}
+
+	const measureName = serialise(team, name, action);
+
+	const options = {
+		start: performance.now(),
+	} satisfies PerformanceMeasureOptions;
+
+	const endPerformanceMeasure = () => {
+		const { duration } = window.performance.measure(measureName, options) ??
+			window.performance.getEntriesByName(measureName, 'measure').at(-1) ?? {
+				duration: performance.now() - options.start,
+			};
+
+		return Math.ceil(duration);
+	};
+
+	return { endPerformanceMeasure };
+};
--- a/libs/@guardian/libs/src/index.test.ts
+++ b/libs/@guardian/libs/src/index.test.ts
@@ -26,6 +26,7 @@ describe('The package', () => {
 			'removeCookie',
 			'setCookie',
 			'setSessionCookie',
+			'startPerformanceMeasure',
 			'storage',
 			'timeAgo',
 		]);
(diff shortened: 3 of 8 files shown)


libs/@guardian/libs/src/measure/@types/measure.ts
```diff
@@ -0,0 +1,19 @@
...
+		 * Firefox returned `undefined` before v101
+		 * @see https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark#browser_compatibility
+		 */
+		measure(
+			measureName: string,
+			startOrMeasureOptions?: string | PerformanceMeasureOptions,
+			endMark?: string,
+		): PerformanceMeasure | undefined;
+	}
+}
```
mxdvl, line 19: Is there a better way to do this? The alternative is to disable [the `no-unnecessary-condition` ESLint rule](https://typescript-eslint.io/rules/no-unnecessary-condition/)

Jakeii (approved):
love it, not sure about the spaces in the separator though, how about `:`, common for KV storage I think?

libs/@guardian/libs/src/measure/measure.ts
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Once ended, measures are appended to the list of `PerformanceMeasure`
+ * and can be retrieved with `performance.getEntriesByType('measure')`
+ *
+ * @example
+ * const stopMeasure = startMeasure('dotcom', 'fetch');
+ * await fetch('https://www.theguardian.com/uk.json');
+ * const duration = stopMeasure();
+ */
+export const startMeasure = (
```
mxdvl, line 24: I wonder what’s the clearest name, here:

- `startMeasure`
- `startPerformanceMeasure`
- `startTimer`
- `start`

cc. @sndrs

libs/@guardian/libs/src/measure/measure.ts
```diff
@@ -0,0 +1,49 @@
...
+): EndMeasure => {
+	if (!('getEntriesByName' in window.performance)) {
+		return fallback;
+	}
+
+	const options = {
+		start: performance.now(),
+	} satisfies PerformanceMeasureOptions;
+
+	return () => {
```
arelra, line 37: Would this mean to stop a measure you would need to have the returned function in the same closure, stored or passed to the point where it is stopped? 

For certain scenarios that span modules or functional boundaries that might be a little cumbersome.

Could we offer another version that allows a start mark name to be passed in to stop the measure?

libs/@guardian/libs/src/measure/measure.ts
arelra: I like aligning the naming to the API so I'd vote for `startMeasure`

libs/@guardian/libs/src/measure/measure.ts
mxdvl: I think that if we needed to cross module boundaries, we could rely in a `PerformanceObserver`. I've got a draft here:

libs/@guardian/libs/src/measure/startMeasure.ts
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * Once ended, measures are appended to the list of `PerformanceMeasure`
+ * and can be retrieved with `performance.getEntriesByType('measure')`
+ *
+ * @example
+ * const { endMeasure } = startMeasure('dotcom', 'fetch');
+ * await fetch('https://www.theguardian.com/uk.json');
+ * const duration = endMeasure();
+ */
+export const startMeasure = (
```
arelra, line 21: Do these functions need to be exported from the main `index.ts`?

mxdvl: Ah yeah I thought about it and then completely forgot!

arelra (approved):
📐

libs/@guardian/libs/src/measure/measure.ts
sndrs: the only thing with this is that 'measure' can refer to more than time (e.g. dimensions). `startPerformanceMeasure` read pretty unambiguously to me?

libs/@guardian/libs/src/performance/startPerformanceMeasure.ts
```diff
@@ -0,0 +1,46 @@
...
+	if (!('getEntriesByName' in window.performance)) {
+		return fallback;
+	}
+
+	const options = {
+		start: performance.now(),
+	} satisfies PerformanceMeasureOptions;
+
+	const endPerformanceMeasure = () => {
+		const measureName = serialise(team, name, action);
```
sndrs, line 35: should we do this in start, before calling `performance.now()`? otherwise the time to serialise will be included in the duration

joecowton1 (approved):

.changeset/spicy-pugs-smell.md
```diff
@@ -0,0 +1,6 @@
+---
+'@guardian/libs': minor
+---
+
+Add the `startMeasure` helper, which has a unified API for measuring durations
```
mxdvl, line 5: This little one escaped me!

mxdvl:
The `measureOptions` attribute of the `performance.measure` is only supported from Safari 14.1

We needed a [hot fix in DCR](https://github.com/guardian/dotcom-rendering/pull/8445).

PR #733, opened 2023-07-21
feat: add collection of INP metrics to core-web-vitals

jamesgorrie:
## What are you changing?

- Replacing the deprecated `ReportHandler` with `ReportCallback`
- Replacing the deprecated `getX` with `onX`
- Adds collection of [INP data](https://web.dev/inp/)

## Why?

`INP` is being used more widely and being reported in tools like Google Search Console (we've recently seen some warnings) and becoming [the replacement for FID in the future](https://web.dev/inp-cwv/). 

Let's get reporting on it so we can respond to it sooner rather than later.

---

## Filesize

I am not 100% sure of how we would know how much filesize this might add to the lib, but given the [source file is quite small](https://github.com/GoogleChrome/web-vitals/blob/3806160ffbc93c3c4abf210a167b81228172b31c/src/onINP.ts) - we should be OK. We will also be notified on the filesize diff when bumping the consumers version.

This isn't a great way to check filesize as I'd rather know at this point that there's an issue to avoid going through the release cycle to receive the feedback. Again though, it _feels_ OK, and if not, we can decide what to do at that point.

Maybe worth thinking about how we add feedback about filesize to libs that will inevitably end up in the browser at this point in the delivery pipeline.

### DIFF
--- a/.changeset/dull-olives-applaud.md
+++ b/.changeset/dull-olives-applaud.md
@@ -1,0 +1,5 @@
+---
+'@guardian/core-web-vitals': minor
+---
+
+Adds collection of INP data in core-web-vitals
--- a/.changeset/serious-yaks-pull.md
+++ b/.changeset/serious-yaks-pull.md
@@ -1,0 +1,5 @@
+---
+'@guardian/core-web-vitals': minor
+---
+
+Updated deprecated ReportHandler with ReportCallback in core-web-vitals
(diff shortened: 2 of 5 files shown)


joecowton1 (approved):

mxdvl (approved):
Let's have some data before it's officially rolled out!

jamesgorrie:
> Regarding filesize, [we're loading this lib dynamically](https://github.com/guardian/csnx/blob/6776c3802ac9047339d45c57dda3a3eadb5e7a47/libs/%40guardian/core-web-vitals/src/index.ts#L89), once we know that the measured data will be used. I think that’s sufficiently accounted for and should not be an issue.

I was just concerned, as we have now added onINP - this would add some filesize, but we haven't got insight into how much. Which in this case is OK, I think we both agree, but wondered how we might account for this in other less controlled situations.

This workflow seems like it could be tedious and might lead to releasing code with more `KB`s that we might tolerate if it wasn't so hard to fix.
1. Add feature to csnx
1. Release
1. bump consumer
1. Find out I've added shed-loads of `KB`s to my JS
1. Back to `1.`

PR #736, opened 2023-07-24
Allow call of `endMeasure` from anywhere

mxdvl:
## What are you changing?

- add a method to `endMeasure` from anywhere

## Why?

- @arelra [suggested it could be worthwhile](https://github.com/guardian/csnx/pull/729#discussion_r1272118294)

### DIFF
--- a/libs/@guardian/libs/src/measure/startMeasure.ts
+++ b/libs/@guardian/libs/src/measure/startMeasure.ts
@@ -5,6 +5,9 @@ import { serialise } from './serialise';
 /** For browser which do not fully support the performance API */
 const fallback: ReturnType<typeof startMeasure> = {
 	endMeasure: () => -1,
+	observe: () => {
+		/* do nothing… */
+	},
 };
 
 /**
@@ -22,7 +25,7 @@ export const startMeasure = (
 	team: TeamName,
 	name: string,
 	action?: string,
-): { endMeasure: () => number } => {
+): { endMeasure: () => number; observe: () => void } => {
 	if (!('getEntriesByName' in window.performance)) {
 		return fallback;
 	}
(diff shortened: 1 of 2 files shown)


libs/@guardian/libs/src/measure/startMeasure.ts
```diff
@@ -31,16 +34,28 @@ export const startMeasure = (
...
-	const endMeasure = () => {
-		const measureName = serialise(team, name, action);
+	const measureName = serialise(team, name, action);
 
-		const { duration } = window.performance.measure(measureName, options) ??
+	const endMeasure = (end?: number) => {
+		const { duration } = window.performance.measure(measureName, {
+			...options,
+			end,
+		}) ??
```
arelra, line 43: Just wondering when the LHS would this return nullish? According to the docs `measure` returns either a `PerformanceMeasure` or throws an exception.

libs/@guardian/libs/src/measure/startMeasure.ts
```diff
@@ -31,16 +34,28 @@ export const startMeasure = (
...
 			window.performance.getEntriesByName(measureName, 'measure').at(-1) ?? {
 				duration: performance.now() - options.start,
 			};
 
 		return Math.ceil(duration);
 	};
 
-	return { endMeasure };
+	const observe = () =>
```
arelra, line 51: I feel like the name could describe what the function does better. 

Just an idea but  `observeEndMark()`?

libs/@guardian/libs/src/measure/startMeasure.ts
```diff
@@ -31,16 +34,28 @@ export const startMeasure = (
...
+		new PerformanceObserver((entries, observer) => {
+			const [endMark] = entries.getEntriesByName(measureName);
```
arelra, line 53: This is cool 😎 . Just a suggestion re semantics and dev tools. Using the zeroth mark as the semantic end mark is a little unusual. I'm wondering if it is less surprising to anyone looking at the performance dev tools to always create start marks. Then we can grab the [1]'st mark as the end mark for `observe`.

libs/@guardian/libs/src/measure/startMeasure.ts
mxdvl: [Firefox before v101 returns `undefined`](https://github.com/guardian/csnx/pull/729/files#diff-d6ff5d412d9e790b174922bb617f2c51fc0df428a74cf235ab0df7c0b1dc6f49R4-R5)

mxdvl:
This seems like a footgun, so let’s wait until there’s an actual need for such a method!

PR #745, opened 2023-07-25
Update @guardian dependencies

OllysCoding:
Updates all packages to use the latest guardian dependencies


libs/@guardian/identity-auth/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # @guardian/identity-auth
 
+## 0.4.0
+
+### Minor Changes
```
sndrs, line 5: if peerDependencies change, [it should trigger a major release](https://github.com/guardian/recommendations/blob/main/npm-packages.md#changes-to-peerdependencies-ranges-are-breaking) rather than a minor one

sndrs (changes requested):
I mentioned the minor/major issue on `@guardian/identity-auth` changelog, but it would apply to most of these i think.

also, given there are no code changes in here, what does bumping these bring consumers? e.g. if i'm happily using `@guardian/identity-auth@0.3.0` with `@guardian/libs@15.2.0` right now, what do i get from these releases?

libs/@guardian/identity-auth/CHANGELOG.md
mxdvl: [Minor version changes before v1 are allowed to have breaking changes](https://semver.org/#spec-item-4)

joecowton1:
> also, given there are no code changes in here, what does bumping these bring consumers? e.g. if i'm happily using `@guardian/identity-auth@0.3.0` with `@guardian/libs@15.2.0` right now, what do i get from these releases?

This is a good point, for context this PR was me demonstrating to James and Olly how we upgrade and release @guardian packages. The value these changes bring to consumers beyond having the latest packages is negligible though. As such maybe we literally don't need this PR..

libs/@guardian/identity-auth/CHANGELOG.md
joecowton1: Thanks Max. We're treating `identity-auth` as pre-release hence why its minor. See comment below on whether its worth doing though.

sndrs: ah yes sorry, the others are all major versions 👍

joecowton1:
Closing in favour of https://github.com/guardian/csnx/pull/753

PR #750, opened 2023-07-27
Update line heights

joecowton1:
## What are you changing?

- updating line height to match changes suggested by the DS team. These will be implemented in Source components and dev kitchen in a future PR.

See here for details: https://github.com/guardian/design-system-advocates/issues/1


jamesmockett (approved):

libs/@guardian/source-foundations/src/typography/data.ts
```diff
@@ -90,8 +90,8 @@ export const fonts = {
 
 export const lineHeights = {
 	tight: 1.15,
-	regular: 1.35,
-	loose: 1.5,
+	regular: 1.3,
+	loose: 1.4,
```
mxdvl, line 94: Worth noting that none of these meet the [WCAG 2.1 AA Success Criterion for Text Spacing](https://www.w3.org/WAI/WCAG21/Understanding/text-spacing)

PR #754, opened 2023-07-27
IdentityAuth | Add global session/identity cookie refresh functionality

coldlink:
## What are you changing?

Adds functionality to automatically renew the user's global Okta session and Identity cookies once every 30 days. This is to keep these 2 sessions in sync, and also to keep them logged in over a longer period of time.

This can be controlled through the new `idCookieSessionRefresh` option. When this is set to `true`, it will attempt to refresh the session based on certain parameters to pass. When `false` it will not attempt to refresh in any scenario. This should be based on the `idCookieRefresh` switch from CAPI.

https://github.com/guardian/frontend/blob/383ce2cd451afb131e3c3aa198233ae689d8fe85/common/app/conf/switches/IdentitySwitches.scala

If enabled this works by checking local storage for the last time the token was refreshed, under the key `identity.lastRefresh`. This is stored as a unix timestamp. If this number is more than 30 days since now, we refresh the session by redirecting the user to the profile subdomains `/signin/refresh` endpoint which handles refreshing the Okta session and the Identity cookies. We also update the `identity.lastRefresh` local storage item to be set to now.

The `/signin/refresh` endpoint will redirect the user to the page that they were on before refreshing once complete. 

## Why?

In frontend migration we found some pre-existing code which handled this refresh functionality. We determined that it still needed to be included in both Frontend and DCR, but was only present in the former. By introducing it to this library, we can handle it in once place, and apply it to both.

https://github.com/guardian/frontend/blob/22c84f55d331cfefd707b27f79adfeecad2a0828/static/src/javascripts/projects/common/modules/identity/cookierefresh.ts#L26-L45

This functionality has been around a while in Frontend, but had been updated since e.g.
https://github.com/guardian/frontend/pull/25996
https://github.com/guardian/frontend/pull/25306
https://github.com/guardian/frontend/pull/25359

This has meant that as more users use DCR instead of Frontend this functionality has been gradually lost, and sessions may be getting out of sync. Hopefully this should fix this scenario too.


coldlink:
Issue tracked here: https://github.com/guardian/frontend/issues/26344

libs/@guardian/identity-auth/src/identityAuth.ts
```diff
@@ -52,6 +55,10 @@ export class IdentityAuth<
 			keysUrl: `${this.#options.issuer}/v1/keys`,
 		};
 
+		// before doing anything else, we check if the user's session should be refreshed based on the idCookieSessionRefresh option
+		const profileUrl = this.#options.issuer.split('/oauth2')[0] as ProfileUrl;
```
mxdvl, line 59: Is this type assertion necessary? Could you have a proper handling of the case where the `profileUrl` is incorrect?

libs/@guardian/identity-auth/src/@types/OAuth.ts
```diff
@@ -13,18 +21,17 @@ import type { AccessToken, CustomClaims, IDToken } from './Token';
...
  * @param maxClockSkew - The maximum time drift in seconds between the client and server, defaults to 300 seconds (5 minutes), based on the default maximum tolerance of the Kerberos protocol
+ * @param idCookieSessionRefresh - If switched on (set to `true`, ), will refresh the okta session and identity cookie (GU_U), if required (once every 30 days), should be based on the id-cookie-refresh (idCookieRefresh) switch if being supplied by theguardian.com, defaults to `false`
  */
 export interface IdentityAuthOptions {
 	clientId: string;
-	issuer: `https://profile.${
-		| 'theguardian'
-		| 'code.dev-theguardian'
-		| 'thegulocal'}.com/oauth2/${string}`;
+	issuer: `${ProfileUrl}/${string}`;
```
mxdvl, line 28: Did you lose the `/oauth2` part of the string?

mxdvl (commented):
Are you sure that this code was not bundled inside the commercial code and therefore executed in DCR all the same?

Also, two mini TS notes.

libs/@guardian/identity-auth/src/@types/OAuth.ts
coldlink: Good spot! Yes I did.

libs/@guardian/identity-auth/src/identityAuth.ts
coldlink: I'll add a type guard for the URL, I used the assertion, as the type for the `issuer` always began with one of the `profile` urls, but when using `split` and getting the first element, typescript only thought it was a string, and not a specific type.

coldlink:
> commercial code

as far as i can tell, the cookie refresh functionality was separate from commercial code

https://github.com/search?q=repo%3Aguardian%2Ffrontend%20cookierefresh&type=code

libs/@guardian/identity-auth/src/identityAuth.ts
mxdvl: Just the [`dotcom-rendering`’s `guard` implementation](https://github.com/guardian/dotcom-rendering/blob/21f02cdf5418b1d4472f4d64d8343a939f536a58/dotcom-rendering/src/lib/guard.ts) here in case…

coldlink: ooh

coldlink: updated to use your amazing type guard, have added you as the author for that commit!

libs/@guardian/identity-auth/src/@types/OAuth.ts
```diff
@@ -1,5 +1,18 @@
...
 import type { AccessToken, CustomClaims, IDToken } from './Token';
 
+/**
+ * Defines the `profile` endpoint URL (Identity/Okta endpoint)
+ */
+export const profileUrls = [
+	'https://profile.theguardian.com',
+	'https://profile.code.dev-theguardian.com',
+	'https://profile.thegulocal.com',
+] as const;
```
mxdvl, line 12:
```suggestion
] as const satisfies ReadonlyArray<`https://profile.${string}`>;
```

libs/@guardian/identity-auth/src/@types/guard.ts
```diff
@@ -0,0 +1,7 @@
+/** A method to create type-guard for string unions */
+export const guard =
+	<T extends readonly unknown[]>(array: T) =>
+	(value: unknown): value is (typeof array)[number] =>
+		array.includes(value);
+
+export type Guard<T> = T extends readonly unknown[] ? T[number] : never;
```
mxdvl, line 7: I think a version that takes in guard function as argument is harder to misuse.

```suggestion
export type Guard<T extends (value: unknown) => value is unknown> = T extends (
	value: unknown,
) => value is infer G
	? G
	: never;
```

coldlink: when i tried this, typescript was always suggesting it was a `never`

Screenshot 2023-07-28 at 13 52 50

coldlink: linting would also return

```
*/csnx/libs/@guardian/identity-auth/src/cookieRefresh.ts
         15:22  error  Invalid type "never" of template literal expression  @typescript-eslint/restrict-template-expressions
         47:8   error  Unsafe assignment of an `any` value                  @typescript-eslint/no-unsafe-assignment
         47:21  error  Unsafe member access [0] on an `any` value           @typescript-eslint/no-unsafe-member-access
         47:21  error  Unsafe call of an `any` typed value                  @typescript-eslint/no-unsafe-call
         52:38  error  Invalid type "never" of template literal expression  @typescript-eslint/restrict-template-expressions
       
       */csnx/libs/@guardian/identity-auth/src/identityAuth.ts
         52:21  error  Invalid type "never" of template literal expression  @typescript-eslint/restrict-template-expressions
         53:17  error  Invalid type "never" of template literal expression  @typescript-eslint/restrict-template-expressions
         54:16  error  Invalid type "never" of template literal expression  @typescript-eslint/restrict-template-expressions
       
       */csnx/libs/@guardian/identity-auth/src/token.ts
         669:21  error  Unsafe member access [0] on an `any` value           @typescript-eslint/no-unsafe-member-access
         669:21  error  Unsafe call of an `any` typed value                  @typescript-eslint/no-unsafe-call
         745:6   error  Invalid type "never" of template literal expression  @typescript-eslint/restrict-template-expressions
```

coldlink: ah

```
export type ProfileUrl = Guard<typeof isProfileUrl>;
```

coldlink: missed your bit around it taking a guard function

mxdvl: Sorry wasn't clear: you're meant to pass your actual type-guard function to it, in this case `isProfileUrl`.

libs/@guardian/identity-auth/src/@types/OAuth.ts
coldlink: ooh satisfies 👀

raphaelkabo (approved):
Looks good to me. 😃

PR #755, opened 2023-07-28
Only publish ES modules in compiled `@guardian` packages

sndrs:
### For all packages
- makes Node 16 the minimum supported version for compiled packages (all Node 16 versions support ES modules)
- adds `@csnx/npm-package:verify` which runs [`publint`](https://publint.dev) against the built output
- releases new versions of affected packages

### For compiled packages

_i.e. for packages built from TS source code_

- removes CommonJS version
- sets `"type": "module"`
- removes `main` and `module` fields, using only `exports`

## Why?

Our current packages are [incorrectly configured](https://publint.dev/@guardian/libs@15.4.0). Fixing this is quite tricky, and publishing CommonJS and ESM together exposes us to the dual package hazard https://nodejs.org/api/packages.html#dual-package-hazard.

We only publish CommonJS because we needed to support old versions of Node. But all current Node versions support ES modules now.

Also, [`@guardian/*` packages are expected to be transpiled by consumers](https://github.com/guardian/recommendations/blob/main/npm-packages.md#using-guardian-npm-packages), so this _should_ not affect us.

### DIFF
--- a/tools/nx-plugins/npm-package/build/set-package-defaults.ts
+++ b/tools/nx-plugins/npm-package/build/set-package-defaults.ts
@@ -4,7 +4,6 @@ import sortPkgJson from 'sort-package-json';
 import type { JsonObject } from 'type-fest';
 import type * as WritePackage from 'write-pkg';
 import type { BuildExecutorOptions } from './schema';
-import type { Entries } from './index';
 
 /**
  * THIS IS KLUDGE #ES_NODE_MODULES
@@ -22,7 +21,7 @@ const esmModuleImport = new Function('specifier', 'return import(specifier)');
  */
 export const setPackageDefaults = async (
 	options: BuildExecutorOptions,
-	entries: Entries | undefined,
+	outputEntry?: string,
 ) => {
 	const { readPackage } = (await esmModuleImport(
 		'read-pkg',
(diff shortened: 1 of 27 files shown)


JamieB-gu:
I've done some experimenting and I think DCR and AR will both build ok with this change 👍

tools/nx-plugins/npm-package/build/set-package-defaults.ts
```diff
@@ -52,6 +51,9 @@ export const setPackageDefaults = async (
 			type: 'git',
 			url: 'git+https://github.com/guardian/csnx.git',
 		},
+		engines: {
+			node: '>=16',
+		},
```
mxdvl, line 56: It’s marked as stable and supported [as early as v12](https://nodejs.org/docs/latest-v12.x/api/esm.html)

sndrs: was only added without a flag in https://nodejs.org/fr/blog/release/v12.17.0. sindrsorhus's [recommendation is to start with node 16](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-move-my-commonjs-project-to-esm) and iirc, there's some complications with node 14 too. 

given [16 is EOL in just over a month from now](https://nodejs.org/en/blog/announcements/nodejs16-eol), this seems ok to me?

sndrs:
> I've done some experimenting using `guardian/libs` and I think DCR and AR will both build ok with this change 👍

nice one @JamieB-gu

mxdvl (approved):

PR #756, opened 2023-07-31
Add gu to performance measure names

arelra:
## What are you changing?

Add the prefix 'gu' to performance measures recorded by libs.

## Why?

It will make it easier to group all of the metrics we care about without having to know the individual team names.

### DIFF
--- a/libs/@guardian/libs/src/performance/serialise.ts
+++ b/libs/@guardian/libs/src/performance/serialise.ts
@@ -5,4 +5,4 @@ import type { TeamName } from '../logger/@types/logger';
 const SEPARATOR = ':';
 
 export const serialise = (team: TeamName, name: string, action?: string) =>
-	[team, name, action].filter(isNonNullable).join(SEPARATOR);
+	['gu', team, name, action].filter(isNonNullable).join(SEPARATOR);
--- a/.changeset/khaki-apricots-mix.md
+++ b/.changeset/khaki-apricots-mix.md
@@ -1,0 +1,5 @@
+---
+'@guardian/libs': minor
+---
+
+Add gu prefix to performance measure name


mxdvl (commented):
I wonder if instead we should create a method to retrieve performance measures for given teams, e.g. with the following signature:

```ts
type GetMeasures = (teams: TeamName[]) => PerformanceMeasure
```

libs/@guardian/libs/src/performance/serialise.ts
```diff
@@ -5,4 +5,4 @@ import type { TeamName } from '../logger/@types/logger';
 const SEPARATOR = ':';
 
 export const serialise = (team: TeamName, name: string, action?: string) =>
-	[team, name, action].filter(isNonNullable).join(SEPARATOR);
+	['gu', team, name, action].filter(isNonNullable).join(SEPARATOR);
```
arelra, line 8: @mxdvl a minor suggestion which I can add if you agree:

I find a dot separator easier to visually scan as it adds lots of whitespace between the tokens.

e.g.

`dotcom:SetAdTargeting:import`

vs

`dotcom.SetAdTargeting.import`

I believe we will be transforming the separator to a BigQuery friendly `_` at transmission time regardless.

arelra:
> I wonder if instead we should create a method to retrieve performance measures for given teams, e.g. with the following signature:
> 
> ```ts
> type GetMeasures = (teams: TeamName[]) => PerformanceMeasure
> ```

Having a method to pull by team name is fine but I think it would also be convenient to pull all the 'gu' metrics in one go rather than having to query (and remember) by team name.

I think we've lacked a higher level platform view of the metrics and how each teams affect each other. Making it easier to query across teams would would help.

Is there a reason why `gu` might be a problem?

libs/@guardian/libs/src/performance/serialise.ts
Jakeii: I'm not too bothered by the separator but I'm not a fan of `.` just to make it clear that these have no relation to JS objects or BigQuery schemas/datasets at first glance.

mxdvl:
> I think we've lacked a higher level platform view of the metrics and how each teams affect each other. Making it easier to query across teams would would help.
> 
> Is there a reason why `gu` might be a problem?

Different teams could be recording performance information that may contain personal data. Any team that does cross-team aggregation then becomes liable to understanding the type of data that is being recorded and its privacy implications.

libs/@guardian/libs/src/performance/serialise.ts
mxdvl: I’d actually pick the separator that the **least** readable, if we want to encourage consistency, and opt for a a deserialiser instead…

```ts
type Deserialiser = (measure: PerformanceMeasure) => { team: TeamName, name: string, action?: string, duration: number, start: number }
// or even just for a name…
type Deserialiser = (name: string) => { team: TeamName, name: string, action?: string }
```

arelra: > I'm not too bothered by the separator but I'm not a fan of `.` just to make it clear that these have no relation to JS objects or BigQuery schemas/datasets at first glance.

I hadn't considered that someone would confuse them with JS objects but I guess its possible.

arelra: > I’d actually pick the separator that the **least** readable, if we want to encourage consistency, and opt for a a deserialiser instead…
> 
> ```ts
> type Deserialiser = (measure: PerformanceMeasure) => { team: TeamName, name: string, action?: string, duration: number, start: number }
> // or even just for a name…
> type Deserialiser = (name: string) => { team: TeamName, name: string, action?: string }
> ```

I tend to look at performance metrics in dev tools quite a bit so this is less readable for me but happy to go with the consensus.

arelra:
> > I think we've lacked a higher level platform view of the metrics and how each teams affect each other. Making it easier to query across teams would would help.
> > Is there a reason why `gu` might be a problem?
> 
> Different teams could be recording performance information that may contain personal data. Any team that does cross-team aggregation then becomes liable to understanding the type of data that is being recorded and its privacy implications.

I was hoping that in any data aggregation pipeline we should always be aware of privacy concerns. You could still aggregate by a single or multiple team names and not be aware of the underlying data. 

I'll close this PR for now though.

PR #767, opened 2023-08-01
Remove storybook-deployer

joecowton1:
## What are you changing?

- test to remove `@storybook/storybook-deployer` 
- replaces with https://github.com/peaceiris/actions-gh-pages

PR #768, opened 2023-08-01
Override `git-url-parse` version to fix critical security vulnerability

jamesmockett:
## What are you changing?

Adds version override for `git-url-parse`

## Why?

`git-url-parse` includes `parse-url` as a sub-dependency which has a critical security vulnerability.

`git-url-parse` itself is a dependency of `@storybook/storybook-deployer` which is now end-of-life so is unlikely to be updated to address this. In the long term, we should replace this package completely.


jamesmockett:
Closing in favour of #767 which removes `@storybook/storybook-deployer` entirely.

PR #773, opened 2023-08-02
Handle performance measures in older browsers

mxdvl:
## What are you changing?

- handle errors in older browsers when passing a [`measureOptions`](https://developer.mozilla.org/en-US/docs/Web/API/Performance/measure#measureoptions) to `window.performance.measure`

## Why?

Some browsers do not support the `measureOptions` parameter to `window.performance.measure()` API introduced in #729, causing the method to throw an error. [According to MDN](https://developer.mozilla.org/en-US/docs/Web/API/Performance/measure#browser_compatibility):

- Chrome below 78 (October 2019)
- Edge below 79 (January 2020)
- Firefox below 103 (May 2022)
- Opera below 65 or 56 (November 2019)
- Safari below 14.1 or 14.5 (April 2021)

Currently addressed by a [hot fix in DCR](https://github.com/guardian/dotcom-rendering/pull/8445).

## Does it work?

Tested in Safari 13.1 using BrowserStack

image

### DIFF
--- a/libs/@guardian/libs/src/performance/startPerformanceMeasure.ts
+++ b/libs/@guardian/libs/src/performance/startPerformanceMeasure.ts
@@ -1,11 +1,16 @@
 import type { TeamName } from '../logger/@types/logger';
+import { log } from '../logger/log';
 import type {} from './@types/measure';
 import { serialise } from './serialise';
 
 /** For browser which do not fully support the performance API */
-const fallback: ReturnType<typeof startPerformanceMeasure> = {
-	endPerformanceMeasure: () => -1,
-};
+const fallbackDuration = -1;
+
+/** return a pseudo random identifier to prevent duplicate mark names */
+const getId = () =>
+	Math.trunc(Math.random() * 1_679_615)
+		.toString(36)
+		.padStart(4, '0');
 
 /**
  * Helper to measure the duration between two events.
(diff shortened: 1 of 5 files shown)


libs/@guardian/libs/src/performance/startPerformanceMeasure.ts
```diff
@@ -27,22 +33,19 @@ export const startPerformanceMeasure = (
...
-		detail: {
-			team,
-			name,
-			action,
-		},
-	} satisfies PerformanceMeasureOptions;
+	// Browser support for `measureOptions` is not good enough,
+	// so we have to rely on the side effect of adding a mark.
+	// See https://developer.mozilla.org/en-US/docs/Web/API/Performance/measure#measureoptions
+	performance.mark(markName);
```
abeddow91, line 43: thanks for the context!

abeddow91 (approved):
Looks good to me, thanks Max 🚀

abeddow91 (approved):

libs/@guardian/libs/src/performance/startPerformanceMeasure.ts
```diff
@@ -23,29 +28,35 @@ export const startPerformanceMeasure = (
...
+		};
 
-		return Math.ceil(duration);
-	};
-
-	return { endPerformanceMeasure };
+		return { endPerformanceMeasure };
+	} catch (error) {
+		log(team, error);
+		return { endPerformanceMeasure: () => fallbackDuration };
```
jamesgorrie, line 60: If we use this to analyse data, would `-1` skew that?

mxdvl: It could, but we're not directly using it yet. I followed the pattern of [`indexOf`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf), [`findIndex`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) and the likes returning **-1**, which may have been ill-advised.

Would `undefined` be more accurate? It would require a major version bump.

PR #779, opened 2023-08-07
Update eslint

joecowton1:
## What are you changing?

- update eslint and other deps

## Why?

- housekeeping


libs/@guardian/eslint-config-typescript/package.json
```diff
@@ -11,7 +11,7 @@
 		"eslint-plugin-import": "2.27.5"
 	},
 	"devDependencies": {
-		"eslint": "8.0.0",
+		"eslint": "8.46.0",
```
mxdvl, line 14: You should not be allowed to do that if your `peerDependencies` specify `8.0.0`.

This would have been caught by #687

PR #781, opened 2023-08-09
Bump Jest packages to fix security vulnerability

jamesmockett:
## What are you changing?

- Bump `jest` and `jest-environment-jsdom` package versions.

## Why?

- This fixes a security vulnerability in `tough-cookie` which is a sub-dependency of `jest-environment-jsdom`.


OllysCoding (approved):
+1 Nice one :)

PR #782, opened 2023-08-10
Update ESLint

jamesmockett:
## What are you changing?

Updates ESLint to `8.47.0`.

## Why?

There is a security vulnerability in the `word-wrap` package which is a dependency of the current version.


mxdvl (commented):
The change in the top `package.json` is 👌!

Do we actually need to bump `@guardian/*` packages? Should it not be the consumers’ responsibility to make sure they do not have vulnerabilities?

mxdvl (approved):

OllysCoding (approved):
+1

PR #783, opened 2023-08-11
Add a new `@guardian/identity-auth-frontend` library which takes care of setting up `@guardian/identity-auth` for DCR, Frontend, and Commerci

AshCorr:
## What are you changing?

Creates a `@guardian/identity-auth-frontend` library which acts as a wrapper around `@guardian/identity-auth` to ensure only a single instance of IdentityAuth is created and that all clients trying to access it use the same settings.

## Why?

- Somewhat ensures that Frontend, DCR, and Commercial all initialize `IdentityAuth` the same way. As they all need to share the same token it would be problematic if one application requested different/less scopes than another application.
- Ensures only 1 instance of IdentityAuth is created per page view. Multiple instances of `IdentityAuth` could potentially try and refresh tokens at the same time as eachother and use up our Okta API requests quota.
- Makes sure that no application is blocked by another. Commercial shouldn't have to wait for DCR/Frontend to initialize to be able to check signed in status.

### DIFF
--- a/.changeset/fuzzy-falcons-build.md
+++ b/.changeset/fuzzy-falcons-build.md
@@ -1,0 +1,5 @@
+---
+'@guardian/identity-auth-frontend': minor
+---
+
+Add a new @guardian/identity-auth-frontend library which takes care of setting up `@guardian/identity-auth` for DCR, Frontend, and Commercial
--- a/.changeset/fuzzy-falcons-build.md
+++ b/.changeset/fuzzy-falcons-build.md
@@ -1,5 +1,0 @@
----
-'@guardian/identity-auth-frontend': minor
----
-
-Add a new @guardian/identity-auth-frontend library which takes care of setting up `@guardian/identity-auth` for DCR, Frontend, and Commercial
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -20,5 +20,6 @@
 /libs/@guardian/eslint-config-typescript/ @guardian/client-side-infra
 /libs/@guardian/browserlist-config/  @guardian/client-side-infra @guardian/dotcom-platform
 /libs/@guardian/libs/ @guardian/client-side-infra @guardian/apps-rendering @guardian/dotcom-platform
+/libs/@guardian/identity-auth-frontend/ @guardian/dotcom-platform @guardian/commercial-dev
 
 /libs/@guardian/identity-auth/ @guardian/identity
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -20,6 +20,6 @@
 /libs/@guardian/eslint-config-typescript/ @guardian/client-side-infra
 /libs/@guardian/browserlist-config/  @guardian/client-side-infra @guardian/dotcom-platform
 /libs/@guardian/libs/ @guardian/client-side-infra @guardian/apps-rendering @guardian/dotcom-platform
-/libs/@guardian/identity-auth-frontend/ @guardian/dotcom-platform @guardian/commercial-dev
+/libs/@guardian/identity-auth-frontend/ @guardian/dotcom-platform @guardian/commercial-dev @guardian/identity
 
 /libs/@guardian/identity-auth/ @guardian/identity
--- a/libs/@guardian/identity-auth-frontend/src/index.ts
+++ b/libs/@guardian/identity-auth-frontend/src/index.ts
@@ -1,0 +1,66 @@
+import { IdentityAuth } from '@guardian/identity-auth';
+import type { CustomClaims } from '@guardian/identity-auth';
+
+type FrontendIdTokenClaims = CustomClaims & {
+	email: string;
+	braze_uuid: string;
+};
+
+export type FrontendIdentityAuth = IdentityAuth<never, FrontendIdTokenClaims>;
+
+const getStage = (isDev: boolean, stage: string) => (isDev ? 'DEV' : stage);
+
+const getIssuer = (stage: string) =>
+	stage === 'PROD'
+		? 'https://profile.theguardian.com/oauth2/aus3xgj525jYQRowl417'
+		: 'https://profile.code.dev-theguardian.com/oauth2/aus3v9gla95Toj0EE0x7';
+
+const getClientId = (stage: string) =>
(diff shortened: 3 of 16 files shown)


libs/@guardian/identity-auth-frontend/src/index.ts
```diff
@@ -0,0 +1,66 @@
...
+	const { config } = guardian ?? {};
+	const { isDev, stage } = guardian?.config ?? {};
+
+	if (!guardian || !config || !isDev || !stage) {
+		throw new Error('window.guardian has not yet been initialized');
+	}
+
+	const stageOrDev = getStage(isDev, stage);
+
+	if (!guardian.identityAuth) {
```
AshCorr, line 44: Instead of saving the instance of IdentityAuth to the window I could have stored it inside the module as a global field, however I was worried that differences between module loading of Frontend vs Commercial might cause the module to be imported/created multiple times and have different instances of IdentityAuth

coldlink (approved):
Approving on behalf of Identity

.github/CODEOWNERS
```diff
@@ -20,5 +20,6 @@
 /libs/@guardian/eslint-config-typescript/ @guardian/client-side-infra
 /libs/@guardian/browserlist-config/  @guardian/client-side-infra @guardian/dotcom-platform
 /libs/@guardian/libs/ @guardian/client-side-infra @guardian/apps-rendering @guardian/dotcom-platform
+/libs/@guardian/identity-auth-frontend/ @guardian/dotcom-platform @guardian/commercial-dev
```
coldlink, line 23:
```suggestion
/libs/@guardian/identity-auth-frontend/ @guardian/dotcom-platform @guardian/commercial-dev @guardian/identity
```

think it might be useful to have identity here too

libs/@guardian/identity-auth-frontend/src/index.ts
mxdvl: We use [conditional assignment with `||=`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR_assignment) in [`@guardian/consent-management-platform`](https://github.com/guardian/consent-management-platform/blob/e768b95da95ded9c6320247cf5ddebb9beeb19a4/src/index.ts#L91-L103)

libs/@guardian/identity-auth-frontend/src/index.ts
```diff
@@ -0,0 +1,66 @@
...
+};
+
+export const getIdentityAuth = () => {
+	const { guardian } = window;
+	const { config } = guardian ?? {};
+	const { isDev, stage } = guardian?.config ?? {};
+
+	if (!guardian || !config || !isDev || !stage) {
+		throw new Error('window.guardian has not yet been initialized');
+	}
```
mxdvl, line 40: Are you expecting this to throw an error in production, when `isDev === false`?


---

FYI Typescript will happily narrow things for you, so you can go step by step with your errors:

```ts
if(!window.guardian) throw new Error('window.guardian has not yet been initialized');
const { config } = window.guardian 
// `config` is always defined here
// etc…
```

AshCorr: > FYI Typescript will happily narrow things for you, so you can go step by step with your errors:

I'm not sure if theres an issue with how Typescript is setup with this project, but typescript was refusing to narrow this for me when I tried!

.changeset/fuzzy-falcons-build.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/identity-auth-frontend': minor
```
jamesmockett, line 2: This will bump the version defined in `package.json` from `0.1.0` to `0.2.0` which might not be what you want as it's the initial release? You can run `pnpm changeset version` to make the version bump and CHANGELOG update locally which you can then tweak as needed and commit.

AshCorr: Ah great! Will this still trigger a release to NPM if I do this locally?

jamesmockett: Yeah, if those are committed it will trigger a release when this PR is merged.

libs/@guardian/identity-auth-frontend/src/index.ts
```diff
@@ -0,0 +1,66 @@
...
+
+const getStage = (isDev: boolean, stage: string) => (isDev ? 'DEV' : stage);
+
+const getIssuer = (stage: string) =>
+	stage === 'PROD'
+		? 'https://profile.theguardian.com/oauth2/aus3xgj525jYQRowl417'
+		: 'https://profile.code.dev-theguardian.com/oauth2/aus3v9gla95Toj0EE0x7';
+
+const getClientId = (stage: string) =>
+	stage === 'PROD' ? '0oa79m1fmgzrtaHc1417' : '0oa53x6k5wGYXOGzm0x7';
```
OllysCoding, line 19: What are these hardcoded (seemingly random) values? 

Are they specific only to the use on `www.theguardian.com` & if so should we be explicit about that in a comment here & perhaps in a documentation section titled `Can I use this on another domain` or something?

AshCorr: They are specifically for use on `www.theguardian.com`. Will add some documentation!

libs/@guardian/identity-auth-frontend/src/index.ts
```diff
@@ -0,0 +1,66 @@
...
+import type { CustomClaims } from '@guardian/identity-auth';
+
+type FrontendIdTokenClaims = CustomClaims & {
+	email: string;
+	braze_uuid: string;
+};
+
+export type FrontendIdentityAuth = IdentityAuth<never, FrontendIdTokenClaims>;
+
+const getStage = (isDev: boolean, stage: string) => (isDev ? 'DEV' : stage);
```
mxdvl, line 11: If you wanted to create a `Stage` type, you could reuse the `guard` helper…

```ts
import type { Guard } from '@guardian/identity-auth/cjs/@types/guard';
import { guard } from '@guardian/identity-auth/cjs/@types/guard';

const stages = ['PROD', 'CODE', 'DEV'] as const;
const isStage = guard(stages);
type Stage = Guard<typeof isStage>;

const getStage = (isDev: boolean, stage: string): Stage => {
	if (isDev) return 'DEV';
	if (isStage(stage)) return stage;
	return 'DEV';
};
```

PR #788, opened 2023-08-14
Identity Auth | Fix clock skew issue, increase OAuth timeout.

coldlink:
### Clock Skew

A bug was found that we were calculating the clock skew every time we were decoding the tokens. While this was fine when getting new tokens from the Okta OAuth API, when decoding tokens on following page views we were using the "now" set to current time the js executed on the new page load. This causes the clockSkew to increase on every page load, as it was being calculated incorrectly. When this got to over 5 mins, an error was thrown.

This PR fixes this issue by calculating the clock skew when the token is received from the API rather than when decoding
tokens. The clock skew is then stored along side the token in local storage so we can use the value again on future page loads.

We also change the function signature of the `Token.decodeTokens` method, so technically this is a breaking change, but will not affect any usage, as currently the only usage is within this
library.

### OAuth Timeout

Updates the default oauth timeout from 12 seconds to 30 seconds as we were seeing a few sentry issues which had timeout errors. Hopefully by increasing the timeout the number of issues will reduce.

An `oauthTimeout` parameter was also added to the options, just in case this needs to be customised on a per application level.


raphaelkabo (approved):
This looks like it'll solve the bug!

PR #680, opened 2023-07-04
Deprecate `@guardian/libs`'s `ArticlePillar` in favour of `Pillar`.

sndrs:
## What are you changing?

- Deprecate `@guardian/libs`'s `ArticlePillar` in favour of `Pillar`

_Co-authored with @joecowton1, supersedes #679._

## Why?

In DCR (one of the two primary consumers of `ArticleFormat`, along with AR) we support several categories of pages that aren't articles: fronts, the all newsletters page etc.. These pages still have a concept of pillar, to use in the nav and footer for example, and need to make use of this `Pillar` type. Therefore we'd like to rename it to reflect its wider usage.

For more context see the following:

- https://github.com/guardian/dotcom-rendering/issues/7562
- https://github.com/guardian/dotcom-rendering/pull/7587#discussion_r1194931855
- https://github.com/guardian/dotcom-rendering/pull/7967

PR #790, opened 2023-08-15
Add `isOneOf`: a literal union predicate helper

mxdvl:
## What are you changing?

- add `isOneOf` helper method, that creates [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) that narrow primitives to [literal](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types) [type unions](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types)

## Why?

- it’s convenient and is already used in [`dotcom-rendering`](https://github.com/guardian/dotcom-rendering/blob/ae8d2ece7fd75fa5ecfbef102d1e7b622622cad2/dotcom-rendering/src/lib/guard.ts) and [`@guardian/identity-auth`](https://github.com/guardian/csnx/blob/ba33113171060eae6164324540ce9ec03033be49/libs/%2540guardian/identity-auth/src/%2540types/guard.ts).

## Example 

```ts
import { isOneOf } from "@guardian/libs";

const stages = ["DEV", "PROD"] as const;
const isStage = isOneOf(stages);

const maybeStage = process.env.STAGE;
if (isStage(maybeStage)) {
  // here maybeStage is narrowed to "DEV" | "PROD"
}

```


libs/@guardian/libs/src/literalUnionPredicate/literalUnionPredicate.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * Create [type predicates][] that narrow primitives
+ * to [literal][] [type unions][].
+ *
+ * [type predicates]: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates
+ * [literal]: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types
+ * [type unions]: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types
+ *
+ * @example
+ * import { createPredicate } from '@guardian/libs';
```
AshCorr, line 10:
```suggestion
 * import { createUnionPredicate } from '@guardian/libs';
```

mxdvl: Ah yeah also need to export it from the index… thanks @AshCorr 🤦

libs/@guardian/libs/src/literalUnionPredicate/literalUnionPredicate.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @example
+ * import { literalUnionPredicate } from '@guardian/libs';
+ *
+ * const stages = ["PROD", "CODE", "DEV"] as const;
+ * const isStage = createPredicate(stages);
+ *
+ * if(!isStage("NOT_A_STAGE")) throw new Error("Invalid stage")
+ */
+export const literalUnionPredicate =
+	<Literals extends readonly unknown[]>(literals: Literals) =>
```
mxdvl, line 18: The only valid literals are `string` and `number`.

```suggestion
	<Literals extends ReadOnlyArray<string | number>>(literals: Literals) =>
```

AshCorr (approved):
Awesome, this will be quite useful!

chrislomaxjones (approved):
Very cool!

mxdvl:
I tried to use the `const` type parameter in a1e60ae399266099a53f82face60d855413e29ba from @chrislomaxjones’s demo in the Client-Side Meetup, but Rollup failed to parse the file…

libs/@guardian/libs/src/index.ts
```diff
@@ -31,6 +31,8 @@ export { isUndefined } from './isUndefined/isUndefined';
 
 export { joinUrl } from './joinUrl/joinUrl';
 
+export { literalUnionPredicate } from './literalUnionPredicate/literalUnionPredicate';
```
sndrs, line 34: i'm not sure how ergonomic/accessible this name will be to people newly learning TS (i had to read the example to see what it did).

given we have things like `isString`, could we use less technical language? e.g. `isOneOf`?

sndrs (changes requested):

libs/@guardian/libs/src/index.ts
mxdvl: I originally went with simply `is`! Naming things is hard and suggestions are welcome and hope to get the best thing before merging. Previous thoughts before I settled on the currently overly explicit `literalUnionPredicate`:

- `is`
- `isLiteral`
- `isOneOf`
- `isIn`
- `guard`

---

Wondering if this requires a change request, though? There’s not been an approval from CSTI yet. Changing names can also be done in a non-breaking ways, as was done for [Pillar](https://github.com/guardian/csnx/pull/680), so the cost of renaming is not unsurmoutable.

sndrs: I'd vote for `isOneOf`, it seems to me to best describe what you've learned about the candidate?

would be good to get a name that seems good before merging. no problem with making changes if it turns out be a bad one, but we should do that because we discovered something, not because we could decide wink 😜

jamesmockett (approved):

libs/@guardian/libs/src/index.ts
mxdvl: `isOneOf` it is!

libs/@guardian/libs/src/index.ts
```diff
@@ -25,6 +25,7 @@ export { Pillar } from './format/Pillar';
 
 export { isBoolean } from './isBoolean/isBoolean';
 export { isNonNullable } from './isNonNullable/isNonNullable';
+export { isOneOf } from './isOneOf/isOneOf';
 export { isObject } from './isObject/isObject';
```
sndrs, line 29:
```suggestion
export { isObject } from './isObject/isObject';
export { isOneOf } from './isOneOf/isOneOf';
```

libs/@guardian/libs/src/isOneOf/README.md
```diff
@@ -0,0 +1,26 @@
+# `isOneOf(literals)`
```
sndrs, line 1: can you add the top-level readme too?

sndrs (approved):

coldlink (approved):

PR #797, opened 2023-08-21
Update ESLint in `@guardian/*` packages

jamesmockett:
## What are you changing?

This matches the update to the root workspace version (#782) and updates all `@guardian/*` packages to `8.47.0`.

## Why?

Older versions of ESLint rely on a vulnerable version of the `word-wrap` package. This ensures we're not publishing packages that include security vulnerabilities and allows individual projects to update to a non-vulnerable version.

There is a mismatch between the minor versions of the `eslint` and `@types/eslint` packages, but this was deemed to be acceptable as this is already the case and fixing a vulnerability takes precedence.


OllysCoding (approved):
+1 :)

PR #804, opened 2023-08-24
Make labels in logs from `logger` less full on

sndrs:
## What are you changing?

updates logs to look like this:

262933845-2b114fa9-ecb6-4133-ab80-b04e65055db6

## Why?

- when there are lot, it's very shouty, and labels dominate over what you've actually logged

### DIFF
--- a/.changeset/dry-brooms-sparkle.md
+++ b/.changeset/dry-brooms-sparkle.md
@@ -1,0 +1,5 @@
+---
+'@guardian/libs': minor
+---
+
+Labels used in `logger` are quieter.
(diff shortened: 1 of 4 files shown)


jamesmockett (approved):
Thanks Alex, this definitely makes the labels feel a lot calmer ☺️

mxdvl (approved):
Very cute! It would be nice to ensure the SVG is visually identical by [changing the radius there, too](https://github.com/guardian/csnx/blob/bf03407b176b3b9cfccf807f7a69f9a2dabfbec6/libs/%40guardian/libs/scripts/generateSvg.logger.teams.ts#L39)?

**EDIT:** fix in #806

PR #809, opened 2023-08-29
Don’t hit `localStorage` for every log in `@guardian/libs#logger`

sndrs:
## What are you changing?

Use a local cache to manage subscriptions in `logger`, instead of hitting `localStorage` on every `log()` call.

## Why?

`localStorage#get` is blocking and slow, and `logger` is only a developer convenience (and it still happens even if the console is closed, so for everyone, all the time).

This is a benchmark using https://github.com/google/tachometer comparing this version with the current one on my Apple M1 Max

image


mxdvl (approved):
LGTM – glad to see the logger being so heavily used that we need to consider its performance :)

libs/@guardian/libs/src/logger/logger.ts
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * Helpers for unit tests.
+ */
+export const _ = {
+	clearSubscriptionsCache: () => {
+		SUBSCRIPTIONS_CACHE = undefined;
+	},
+	subscribeTo,
+	unsubscribeFrom,
```
mxdvl, line 124: Are these not available on `window.guardian`?

sndrs: not right now https://github.com/guardian/csnx/blob/8a34fe9783c06e426fb765c84519ff52667dce2a/libs/%40guardian/libs/src/logger/log.ts#L55-L65

OllysCoding (approved):
+1 LGTM, nice work :)

PR #813, opened 2023-09-07
Dblatcher/newsletter types in libs

dblatcher:
## What are you changing?

- exports a set of types from @guardian/libs describing the output of the [newsletters api](https://newsletters.guardianapis.com/newsletters), published by the [newsletters-nx](https://github.com/guardian/newsletters-nx) project.

## Why?

- avoid duplicating the definitions of the Newsletters Api data in multiple TypeScript projects consuming data from the API
- include annotations explaining the fields

### DIFF
--- a/.changeset/heavy-rings-visit.md
+++ b/.changeset/heavy-rings-visit.md
@@ -1,0 +1,5 @@
+---
+'@guardian/libs': minor
+---
+
+add types for objects from the newsletters api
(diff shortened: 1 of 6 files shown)


OllysCoding:
@dblatcher I'm not sure if this is best placed in libs (which I mostly consider to be a home of utils), maybe we could have a new package, `@guardian/newsletter-types` or something?

I realise that the inclusion of `ophan` types in libs does contradict this thought, but I think that ophan types also seem out of place in libs ...

dblatcher:
using a new project rather than adding to libs: https://github.com/guardian/csnx/pull/831

PR #831, opened 2023-10-03
newsletters type package

dblatcher:
## What are you changing?

- add a new package project that exports types describing the output of the [newsletters API](https://github.com/guardian/newsletters-nx)

## Why?

- provides shared definitions that ts projects using the API can import instead of defining their own types
- originally suggested adding to @guardian/libs , but using a separate package should be more flexible


joecowton1:
The new project will need adding to the paths in the base tsconfig file too

joecowton1:
LGTM.

joecowton1 (approved):

PR #832, opened 2023-10-03
Update NX packages and fix issue with identity-auth

joecowton1:
## What are you changing?

- add @guardian/identity-auth to root tsconfig paths
- updating nx packages

## Why?

- nx needed updating, updating the pnpm lockfile exposed an issue with identity auth not being found in identity-auth-frontend


OllysCoding (approved):
+1

PR #841, opened 2023-10-09
Remove AudioAtom

joecowton1:
## What are you changing?

- Removing the AudioAtom from atoms-rendering

## Why?

- we recently migrated it to DCR


OllysCoding (approved):
+1

PR #848, opened 2023-10-11
Remove video atom

joecowton1:
## What are you changing?

- remove video atom from atoms rendering

## Why?

- we're migrating atoms to dcr


joecowton1:
@jamesmockett should be fixed now

jamesmockett (approved):

PR #850, opened 2023-10-12
Refactor `timeAgo` to make it more performant

mxdvl:
- feat(duration): expose time internals
- test(duration): check various values

## What are you changing?

-

## Why?

-


jamesgorrie (approved):
Consider that yak shaved!
🐑 ✂️

libs/@guardian/libs/src/datetime/timeAgo.ts
```diff
@@ -1,47 +1,39 @@
-type Unit = 's' | 'm' | 'h' | 'd';
+const units = {
+	second: 1_000,
+	minute: 60_000,
+	hour: 3_600_000,
+	day: 86_400_000,
+} as const satisfies Record<string, number>;
```
abeddow91, line 6: 😍

abeddow91 (approved):

jamesgorrie:
To clarify yak shaving, I meant in a good way:

"An activity that allows one to overcome intermediate difficulties, which then allows one to solve a larger problem."

As in avoiding death of a thousand papercuts on performance.

PR #851, opened 2023-10-12
Enable `timeAgo` to receive a `now` method

mxdvl:
## What are you changing?

- allow passing a `now` optional parameter to `timeAgo`

## Why?

- this allows to use a different time that the one at which the method is run, which is helpful for
  - matching the client and server clocks to prevent hydration mismatches 
  - avoiding mocks and side-effects – this method can be a pure function

### DIFF
--- a/.changeset/quiet-buttons-move.md
+++ b/.changeset/quiet-buttons-move.md
@@ -1,0 +1,5 @@
+---
+'@guardian/libs': minor
+---
+
+Enable passing `now` as an argument to `timeAgo`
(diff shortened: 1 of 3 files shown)


.changeset/quiet-buttons-move.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/libs': minor
+---
+
+Enable passing `now` as an argument to `timeAgo`
```
sndrs, line 5:
```suggestion
Add a `now` option to `timeAgo`. Useful for situations in which you need to recreate a previous call, for example when hydrating a server-rendered component which uses it.
```

sndrs (approved):
one suggestion for the changelog but great apart from that

abeddow91 (approved):

PR #868, opened 2023-10-18
Removed atoms rendering from csnx

oliverabrahams:
## What are you changing?

- Removed atoms rendering from csnx repo

## Why?

-  Deprecating package. Moved all atoms into DCR


joecowton1 (approved):
Nicely done 👌

PR #892, opened 2023-11-06
Update spacing tokens

jamesmockett:
## What are you changing?

Updates Source spacing token names and adds missing values.

## Why?

The current set of spacing tokens is missing values that have been subsequently added and does not follow the naming conventions used in Figma.

## Additional information

The token names in Figma follow the convention of `spacing.010`, `spacing.020` … `spacing.160`. As these cannot be represented in JS in this exact form the tokens are exported as individual variables with the full stop removed. eg. `spacing010`.

This breaks with the existing convention of exporting tokens as an object with a numeric or alphanumeric index. ie. `spacing010` could instead be represented as `spacing[10]` or `spacing['010']`.

There is benefit in terms of developer ergonomics when using individual variables as VS Code can autocomplete these and show all the available options when you start typing `spacing`. This is not the case when using a numeric index on an object.

Screenshot 2023-11-06 at 17 22 06

You do, however, lose the ability to iterate over the spacing values. This could perhaps be solved by also exporting the individual tokens as an object, although it's unlikely a consumer of the tokens would need access to this.

```js
export const spacings = {
  spacing010,
  spacing020,
  // etc.
}
```

Exporting as individual variables also makes imports more verbose, although it's unlikely that the complete set of tokens will be used in a single place.

```js
import { spacing030, spacing060 } from '@guardian/source-foundations';
```

### Codemod

The existing tokens have been marked as deprecated so are still available to use. To ease migration we should add a codemod to the [`eslint-plugin-source-foundations`](https://github.com/guardian/csnx/tree/main/libs/%40guardian/eslint-plugin-source-foundations`) to automatically update instances of the old tokens before we remove them completely in a later release.

### Spacing units

Currently we make the spacing tokens available in `px` and `rem` units. There is an argument that we should discourage the use of `px` and make `rem` the default, but it's not clear what impact this might have on projects if we were to do so.  There's also an implicit assumption that the base font size is `16px`. (Unless the user has changed their browser settings in which case we should respect that.)

We could provide a `remToPx` helper to enable usage of `px` values if required. eg.:

```jsx
margin-top: ${remToPx(spacing060)}px;
```

### Other tokens

Does it make sense to reorganise other tokens in the same way so that we export as individual variables rather than objects?

```js
export const colorBrand500 = '#0077b6';
export const colorBrand600 = '#506991';
// etc.
```

As colours have a hierarchy they lend themselves more to being organised as an object, unlike something like spacing which has a flat hierarchy.

```js
export const colors = {
  brand500: '#0077b6',
  brand600: '#506991',
  // etc.
};
```


jamesmockett:
Closing this whilst we have a deeper think about tokens and naming.

PR #926, opened 2023-11-15
Format `@guardian/libs` with Prettier

mxdvl:
## What are you changing?

- apply prettier formatting to `@guardian/libs`

## Why?

- reduce the number of inadvertend diff
- Redo of #852

### DIFF
--- a/libs/@guardian/libs/src/cookies/README.md
+++ b/libs/@guardian/libs/src/cookies/README.md
@@ -6,19 +6,19 @@ Robust API over `document.cookie`.
 
 ```js
 import {
-    getCookie,
-    removeCookie,
-    setCookie,
-    setSessionCookie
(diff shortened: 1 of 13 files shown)


sndrs:
hmmm, how did this happen? the files should be formatted on commit, but maybe we need to add a DCR-style check in CI too

mxdvl:
Yes, files are formatted on commit, but they are not formatted in CI.

I could try and address formatting in CI, but I tried to keep this small and that it could be a fix-forward.

sndrs:
> I could try and address formatting in CI, but I tried to keep this small and that it could be a fix-forward.

yeah totally. 

having just done this in dcr, i was ready with the necessities :) #931 should fix these and the others, and adds checks

thanks very much for spotting this

PR #940, opened 2023-11-20
Add Button option to Footer Links

GHaberis:
## What are you changing?

**Objective:** Currently `FooterLinks` only renders anchor tags (`Link` components) that take users to a URL when clicked on. We'd like to update `FooterLinks` so it can also render an item that when clicked on instead executes an `onClick` function. 

The immediate use case is the Support site requires a "Privacy Settings" link in the footer that launches the Guardian's CMP. Visually this link should look the same as the other links, and it could be placed anywhere within the list of link items.

**Implementation:**  Update the `links` prop on `FooterLinks`  to be an array of `FooterLink` _or_ `FooterButton` objects. When rendering `FooterLinks` iterate through the `links` array and render a `Link` if the object in the `links` array is a `FooterLink`, else render a `ButtonLink` as the object in the `links` array will be a `FooterButton`.

The solution uses a button (using Source's `ButtonLink`) instead of an anchor with a click handler for [accessibility](https://github.com/bradbirdsallCHANGED/eslint-plugin-jsx-a11y/blob/main/docs/rules/anchor-is-valid.md) reasons.

**Storybook screenshot (see "Footer Links With Footer Button"):**

Screenshot 2023-11-20 at 17 58 49

### DIFF
--- a/libs/@guardian/source-react-components-development-kitchen/src/footer-with-contents/FooterLinks.tsx
+++ b/libs/@guardian/source-react-components-development-kitchen/src/footer-with-contents/FooterLinks.tsx
@@ -1,25 +1,28 @@
 import { ThemeProvider } from '@emotion/react';
 import type { EmotionJSX } from '@emotion/react/types/jsx-namespace';
 import {
+	Button,
+	buttonThemeBrand,
 	Column,
 	Columns,
 	Hide,
 	Link,
 	linkThemeBrand,
 } from '@guardian/source-react-components';
-import type { AnchorHTMLAttributes } from 'react';
+import type { AnchorHTMLAttributes, ButtonHTMLAttributes } from 'react';
 import {
 	fullWidthContainer,
 	getItemStyles,
 	getListStyles,
 	linkElementStyles,
+	footerButtonColumnItemStyles,
 } from './footerLinksStyles';
 
 type FooterLink = AnchorHTMLAttributes<HTMLAnchorElement> & {
 	/**
 	 * The URL of the link
 	 */
-	href: string;
+	href?: string;
 	/**
 	 * The visible text of the link
 	 */
@@ -30,7 +33,18 @@ type FooterLink = AnchorHTMLAttributes<HTMLAnchorElement> & {
 	isExternal?: boolean;
 };
 
-export const defaultGuardianLinks: FooterLink[] = [
+type FooterButton = ButtonHTMLAttributes<HTMLButtonElement> & {
+	/**
+	 * onClick handler for the link
+	 */
+	onClick: () => void;
+	/**
+	 * The visible text of the link
+	 */
+	text: string;
+};
+
+export const defaultGuardianLinks: Array<FooterLink | FooterButton> = [
 	{
 		href: 'https://www.theguardian.com/info/privacy',
 		text: 'Privacy policy',
@@ -52,7 +66,7 @@ export interface FooterLinksProps {
 	/**
 	 * An array of links, specifying the link text and href, if the link is external, and any other HTML attributes for an anchor tag
 	 */
-	links?: FooterLink[];
+	links?: Array<FooterLink | FooterButton>;
 	/**
 	 * Force the links into a column layout below desktop, regardless of the amount of links
 	 */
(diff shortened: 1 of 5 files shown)


libs/@guardian/source-react-components-development-kitchen/src/footer-with-contents/FooterLinks.tsx
```diff
@@ -19,7 +20,7 @@ type FooterLink = AnchorHTMLAttributes<HTMLAnchorElement> & {
 	/**
 	 * The URL of the link
 	 */
-	href: string;
+	href?: string;
```
jamesgorrie, line 23: Given this is exclusively for links - and we use `<button>`s for non-`href`ed elements, why is this optional?

rBangay, line 23: Does this need to be optional if the FooterLink and FooterButton are two distinct types wouldn't href always be present in a FooterLink?

jamesgorrie: This would also make the [type-guard](https://github.com/guardian/csnx/pull/940/files#diff-4e786fe5e22a0794f76d2b114359a2cd833c90407597c3a55e6e8417107ec951R82) not completely true, because `FooterLink`s can have `.href === undefined`, so in essence the guardian and type don't match.

libs/@guardian/source-react-components-development-kitchen/src/footer-with-contents/FooterLinks.tsx
```diff
@@ -64,46 +76,68 @@ export const FooterLinks = ({
 	forceColumns = false,
 }: FooterLinksProps): EmotionJSX.Element => {
 	const useColumns = links.length > 3 || forceColumns;
+	const isFooterLink = (
+		link: FooterLink | FooterButton,
+	): link is FooterLink => {
+		return (link as FooterLink).href !== undefined;
```
jamesgorrie, line 82: I _think_ you could use something like `href in link` without using type coercions.

e.g.
```ts
type A = { href: string, text: string }
type B = { text: string }

type C = A | B

function isA(c: C): c is A {
    return 'href' in c
}


const b = { text: 'Yo!' }
const bIsAnA = isA(b);
```

libs/@guardian/source-react-components-development-kitchen/src/footer-with-contents/FooterLinks.tsx
```diff
@@ -64,46 +76,68 @@ export const FooterLinks = ({
...
+	};
+	const getLink = (link: FooterLink) => {
```
jamesgorrie, line 84: Is it worth defining these as component outside of FooterLinks as:
* They have the shape of components e.g. `Props => Jsx.Element`
* they aren't dependant on any values from `FooterLinks`
* Avoid these being generated each time `FooterLinks` is instantiated

jamesgorrie (commented):
A few questions, but looks good.

libs/@guardian/source-react-components-development-kitchen/src/footer-with-contents/FooterLinks.tsx
GHaberis: @jamesgorrie @rBangay this doesn't need to be optional, I'd switched it to optional when I was initially looking at the types to see if I could / should use the same type, which I opted not to do, so yes I just forgoto to undo this change! Will restore as non-optional now.

GHaberis: @rBangay see comment above https://github.com/guardian/csnx/pull/940/files#r1400352844

libs/@guardian/source-react-components-development-kitchen/src/footer-with-contents/FooterLinks.tsx
GHaberis: Yeah they don't need to be defined in the scope of `FooterLinks`, moving them out makes it more readable too IMO 👍

libs/@guardian/source-react-components-development-kitchen/src/footer-with-contents/FooterLinks.tsx
```diff
@@ -52,58 +64,82 @@ export interface FooterLinksProps {
...
-								</Link>
-							</ThemeProvider>
-						</li>
-					))}
+					{links.map((link, index) => {
+						{
+							return (
+								<li key={`link-${index}`} css={getItemStyles(useColumns)}>
+									<ThemeProvider theme={linkThemeBrand}>
+										{isFooterLink(link) ? getLink(link) : getButtonLink(link)}
```
mxdvl, line 120: There’s a danger in using type predicates that do not have 100% coverage, so I would probably rely on TypeScript to do the narrowing for you inline:

```suggestion
										{'href' in link ? getLink(link) : getButtonLink(link)}
```

Alternatively a test proving that there is no way to construct an object that would incorrectly interpret a link or button would be very useful!

sookburt (commented):
I like what you're doing here.
If we can't enforce it owning to the dependency, I wonder if something could be added to the README to indicate that, if the consent-management-platform is required on the page, that one of the additions needs to be the Privacy Settings link.
I'm happy to add that if it helps.

GHaberis:
@jamesgorrie @mxdvl @rBangay thanks for feedbacl, I've applied the suggested changes

libs/@guardian/source-react-components-development-kitchen/src/footer-with-contents/FooterLinks.tsx
```diff
@@ -52,58 +64,78 @@ export interface FooterLinksProps {
...
-	links?: FooterLink[];
+	links?: Array<FooterLink | FooterButton>;
 	/**
 	 * Force the links into a column layout below desktop, regardless of the amount of links
 	 */
 	forceColumns?: boolean;
 }
 
+const getLink = (link: FooterLink) => {
+	const { href, text, isExternal, ...linkAttrs } = link;
```
jamesgorrie, line 75: Complete sideline, you could deconstruct in the arguments

```suggestion
const getLink = ({ href, text, isExternal, ...linkAttrs }: FooterLink) => {
```

jamesgorrie (approved):

sndrs (approved):

jamesmockett (approved):

paul-daniel-dempsey (commented):
Much tidier and has footerbutton, perfect, reviewed as part of the support checkout modification.

PR #945, opened 2023-11-20
Update @typescript-eslint versions

joecowton1:
## What are you changing?

`@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser` dependencies upgraded to next major version (6).

## Why?

- housekeeping


.changeset/little-seas-fail.md
```diff
@@ -0,0 +1,8 @@
+---
+'@guardian/eslint-plugin-source-react-components': major
+'@guardian/eslint-plugin-source-foundations': major
+'@guardian/eslint-config-typescript': major
+'@guardian/eslint-config': major
+---
+
+Update eslint version
```
sndrs, line 8:
```suggestion
`@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser` dependencies upgraded to next major version (6).
```

libs/@guardian/eslint-config-typescript/package.json
```diff
@@ -5,18 +5,18 @@
...
-		"eslint-import-resolver-typescript": "3.5.5",
-		"eslint-plugin-import": "2.27.5"
+		"@typescript-eslint/eslint-plugin": "6.12.0",
+		"@typescript-eslint/parser": "6.12.0",
+		"eslint-import-resolver-typescript": "3.6.1",
+		"eslint-plugin-import": "2.29.0"
 	},
 	"devDependencies": {
-		"eslint": "8.47.0",
+		"eslint": "8.54.0",
```
sndrs, line 14: is this bump needed?

joecowton1: no - removed

libs/@guardian/source-react-components/src/button/styles.ts
```diff
@@ -303,8 +303,8 @@ export const buttonStyles =
 		button,
 		sizes[size],
 		priorities[priority](theme.button),
-		iconSvg || isLoading ? iconSizes[size] : '',
-		(iconSvg || isLoading) && !hideLabel ? iconSides[iconSide] : '',
```
oliverabrahams, line 307: if iconSvg is `false` or `0` or `''` and isLoading is `true` this will still evaluate to false now

libs/@guardian/source-react-components/src/button/styles.ts
```diff
@@ -303,8 +303,8 @@ export const buttonStyles =
...
+		iconSvg ?? isLoading ? iconSizes[size] : '',
+		(iconSvg ?? isLoading) && !hideLabel ? iconSides[iconSide] : '',
```
oliverabrahams, line 307: if iconSvg is `false`, `0` or `""`  ```iconSvg ?? isLoading``` will now evaluate as false

oliverabrahams (approved):

libs/@guardian/source-react-components/src/button/styles.ts
joecowton1: I _think_ thats what we want right? This was the linters auto fix iirc

oliverabrahams: Yea i think it doesn't matter because iconSvg has type `iconSvg: React.ReactElement<any, string | React.JSXElementConstructor<any>> | undefined` so it can't be any of those values

joecowton1: my hunch is that we were just using `||` (or now `??`) as a catch all for any falsy value

PR #948, opened 2023-11-23
Make members of AB Ophan API required

mxdvl:
## What are you changing?

Members of the `OphanAPIConfig` are now **required**. Consumers must pass a record function, an error reporter and an object of server-side tests.

In order to update your code to work identically, you must now provide the
following keys to the constructor’s argument. They are listed here along with
the fallbacks previously applied:

- `serverSideTests` &rarr; `{}`
- `errorReporter` &rarr; `() => undefined`
- `ophanRecord` &rarr; `() => undefined`

## Why?

With the previously optional members, failure to record tests to Ophan would fail silently. [This has been the case in `dotcom-rendering` since Oct 2023](https://github.com/guardian/dotcom-rendering/pull/8965/files#r1403449259).

### DIFF
--- a/libs/@guardian/ab-core/src/@types/index.ts
+++ b/libs/@guardian/ab-core/src/@types/index.ts
@@ -23,9 +23,9 @@ export type CoreAPI = {
 };
 
 export type OphanAPIConfig = {
-	serverSideTests?: ServerSideTests;
-	errorReporter?: ErrorReporterFunc;
-	ophanRecord?: OphanRecordFunction;
+	serverSideTests: ServerSideTests;
+	errorReporter: ErrorReporterFunc;
+	ophanRecord: OphanRecordFunction;
 };
 
 export type OphanAPI = {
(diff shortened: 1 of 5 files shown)


libs/@guardian/ab-core/src/@types/index.ts
```diff
@@ -23,9 +23,9 @@ export type CoreAPI = {
 };
 
 export type OphanAPIConfig = {
-	serverSideTests?: ServerSideTests;
-	errorReporter?: ErrorReporterFunc;
-	ophanRecord?: OphanRecordFunction;
+	serverSideTests: ServerSideTests;
```
lindseydew, line 26: Would we want to keep this one as optional? Feels like less of an issue if the object wasn't passed in?

lindseydew (approved):
LGTM, just a question about the serverside tests field

libs/@guardian/ab-core/src/@types/index.ts
mxdvl: We can always make it optional in the future without requiring a major/breaking change.

I think that this library is meant to capture all tests, and part of the reason we did not notice that it had broken was because it’s currently not responsible for registering the tests in `dotcom-rendering`. I think this change will make things more explicit, and can be trivially bypassed with an empty object `{}`.

mxdvl:
@guardian/client-side-infra do I need to bump `@guardian/ab-react` as part of this change?

sndrs:
> @guardian/client-side-infra do I need to bump `@guardian/ab-react` as part of this change?

i think so yes, i think you'll want to bump the peerdependcy in there?

oliverabrahams (approved):
LGTM

PR #951, opened 2023-11-24
First pass at a source foundations stylesheet

jamesmockett:
## What are you changing?

Adds first pass at design tokens for Source with custom transforms for Style Dictionary to generate a base stylesheet.


jamesmockett:
Closing as the hackday stylesheet was only intended to be temporary and as of #976 we have moved from Style Dictionary to Cobalt and DTFM tokens.

PR #963, opened 2023-11-27
Support `ExpandingWrapper` theming with pure CSS

mxdvl:
## What are you changing?

- Refactor ExpandingWrapper so it receives an optional theme object which is used to define relevant custom properties which drive the colour of the component
- Remove CSS Overrides as we enable consumers to override every single colour
- Add stories to demonstrate usage

## Why?

- This enables us to use Callouts with dark mode in dotcom-rendering: https://github.com/guardian/dotcom-rendering/issues/9277
- Proposal for an answer to: https://github.com/guardian/dotcom-rendering/issues/9333

### DIFF
--- a/libs/@guardian/source-react-components-development-kitchen/src/expanding-wrapper/types.ts
+++ b/libs/@guardian/source-react-components-development-kitchen/src/expanding-wrapper/types.ts
@@ -1,5 +1,5 @@
-import type { SerializedStyles } from '@emotion/react';
 import type { ReactElement } from 'react';
+import type { Theme } from './theme';
 
 export interface ExpandingWrapperProps {
 	children: ReactElement;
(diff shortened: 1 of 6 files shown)


oliverabrahams (approved):
LGTM

libs/@guardian/source-react-components-development-kitchen/src/expanding-wrapper/types.ts
```diff
@@ -8,10 +8,7 @@ export interface ExpandingWrapperProps {
 
 	renderExtra?: () => ReactElement;
 	expandCallback?: (expanded: boolean) => void;
-	/**
-	 * Override component styles by passing in the result of [emotion's `css` function/prop](https://emotion.sh/docs/introduction).
-	 */
-	cssOverrides?: SerializedStyles | SerializedStyles[];
+	theme?: Theme;
```
alinaboghiu, line 11: this is neat :)

PR #1007, opened 2023-12-12
restructured the design tokens package.

oliverabrahams:
Co-authored-by: jamesmockett <jamesmockett@users.noreply.github.com>

## What are you changing?

Moved the tokens to be built in the ./tokens directory. Moved the tokens into the same file (tokens.json)

## Why?

By moving the tokens to the ./tokens directory and consolidating them into tokens.json, we align our structure with the Cobalt documentation's guidelines. 

Previously, there was some confusion regarding the relationship between source files and their auto-generated counterparts. This restructuring should clarify that the tokens.json file is the primary source, from which all other token files are generated. It makes it clearer that the auto-generated files are direct outputs of the tokens.json source, reducing  confusion about the origin of our design tokens.


jamesmockett (approved):
👍

PR #1009, opened 2023-12-12
Consume spacing design tokens in Source Foundations

oliverabrahams:
## What are you changing?

- Adding design tokens as the source of truth for spacing in Source Foundations

Co-authored-by: jamesmockett jamesmockett@users.noreply.github.com


libs/@guardian/source-foundations/src/space/space.ts
```diff
@@ -6,79 +7,19 @@ import { pxToRem } from '../utils/px-to-rem';
...
-	16: `${pxToRem(space[16])}rem`,
-	18: `${pxToRem(space[18])}rem`,
-	24: `${pxToRem(space[24])}rem`,
-};
+type SpaceTokenKey = keyof typeof tokens.space;
+
+const space = {} as { [K in SpaceTokenKey]: number };
+const remSpace = {} as { [K in SpaceTokenKey]: string };
+
+Object.keys(tokens.space).forEach((key) => {
```
sndrs, line 17: one aspect of looping here is that some processing now has to be done when the module is loaded, not just parsing the values as before... what do you feel about hard coding, but still consuming the design tokens?

jamesmockett: Oh, that's a good point. Maybe hardcoding it isn't so bad as it's an interim step until we consume the token output directly. I _do_ like that this automatically handles the addition of new spacing tokens, but that's not something that happens that often. (We've recently filled in a lot of the gaps anyway.) We can try it and see how it looks…

oliverabrahams: It would make the types easier also. I am happy to hard code it but we do lose that a change permeates from  figma-> design-tokens -> source foundations without us having to write code

sndrs: we could get round that with a script too? i.e. auto-generate (some of) the source code for foundations?

but that feels like it would be subsequent improvement

sndrs: also gets round needing to run `pxStringToNumber` etc when you import it. there's no value to consumers in running that stuff each time, be great to remove the need (doesn't need to be part of this either)

oliverabrahams: Yea, that would be great. If we could just build the hardcoded values from the design tokens so it doesn't have to go through all this processing when built

sndrs (approved):

PR #1011, opened 2023-12-14
enable running make for any nx target

sndrs:
## What are you changing?

- enables running Nx targets with `make`, e.g. `make @guardian/libs:build`
- lists the available Nx targets in the readme
- adds `make ls` target to list available targets 

## Why?

- make it easier to run individual project tasks

## Screenshots

image

_actual list is longer than this!_


Makefile
```diff
@@ -97,7 +105,14 @@ check-node-version:
 # Install dependencies. If deps are up to date this is almost instant, so we can
 # run before every other target with very little DX cost.
 .PHONY: install # PRIVATE
-install: check-node-version
+install: | check-node-version
```
oliverabrahams, line 108: This is cool! How does this syntax work?

oliverabrahams (approved):
LGTM

Makefile
sndrs: this make the prerequisites [order-only](https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html). 

we actually only need them for the `%` target, because it maintains the value of `$@` after the prerequisites have run. without it, the value changes to `"makefile"`, which i'm _guessing_ is because the prerequisites have no actual output, though i'm not 100% sure. i know this makes it work though!

i've removed them now from the other targets, as they don't have an effect

PR #1036, opened 2023-12-19
Add missing entries to changelogs

joecowton1:
## What are you changing?

- adding missing entries to changelogs
- removing out of date changesets


oliverabrahams (approved):
+1

PR #1050, opened 2024-01-02
Update tokens with missing colour and fix `Breakpoint` type

jamesmockett:
## What are you changing?

- Adds `neutral.38` to colour tokens.
- Revert changes to `Breakpoint` type so we explicitly list out allowed breakpoints.

## Why?

- `neutral.38` was missed when the colour tokens were created as it was only recently added to Source Foundations.
- Using `keyof typeof tokens.breakpoint` to get valid breakpoint names is not bundled correctly as it attempts to import `@csnx/design-tokens` in the type declaration which cannot be found as it is a private package that has not been published. This will be fixed separately, but for now we're reverting to explicitly listing these out as before.


sndrs:
should this be a patch? it's not adding new functionality iiuc (just fixing accidentally removed things)

jamesmockett:
> should this be a patch? it's not adding new functionality iiuc (just fixing accidentally removed things)

Good point, I've downgraded it to a patch release.

sndrs (approved):

PR #1057, opened 2024-01-04
Test Source Foundations token output

jamesmockett:
## What are you changing?

- Adds tests to verify output of Source Foundations tokens and typography API.
- Adds missing fallbacks to font stacks to match previous output before `@csnx/design-tokens` was imported.

## Why?

- This provides an additional safety net to flag changes to the token output to help catch instances of tokens values that have been accidentally changed, or tokens that have been removed or missed.
- The expected output was generated by reverting to an older version of foundations, prior to `@csnx/design-token` becoming the source of truth, and generating a set of snapshot tests which have now been inlined. This highlighted a couple of missing fallbacks in the font stacks which have now been restored.


joecowton1 (approved):
Nice one James, I think this a good approach.

PR #1068, opened 2024-01-08
Clean up typography CSS output

jamesmockett:
## What are you changing?

- Strips out comment from typography CSS and places semicolons in correct place so they're not rendered if the corresponding property has no value and is not output
- Adds custom Jest matcher to normalise CSS with [Lightning CSS](https://lightningcss.dev/) and match the result

## Why?

- Our snapshot tests are more verbose than they need to be due to the unnecessary output from the typography API
- By normalising the expected and received CSS with a custom matcher we don't need to worry about formatting and whitespace in the tests and can concentrate on the correct properties and values being present

### DIFF
--- a/libs/@guardian/source-foundations/src/typography/font-styles.ts
+++ b/libs/@guardian/source-foundations/src/typography/font-styles.ts
@@ -130,16 +130,8 @@ export const fontStyleToStringFunction =
 			font-family: ${fontFamily};
 			font-size: ${typeof fontSize === 'number' ? `${fontSize}px` : fontSize};
 			line-height: ${lineHeight};
-			${fontWeight ? `font-weight: ${fontWeight}` : ''};
-			${fontStyle ? `font-style: ${fontStyle}` : ''};
-
-			/*
-			 * Child elements (e.g. <Link/>) can use this variable
-			 * to set the thickness of their underline.
-			 *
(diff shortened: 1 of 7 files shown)


libs/@guardian/source-foundations/src/typography/font-styles.ts
```diff
@@ -130,16 +130,8 @@ export const fontStyleToStringFunction =
...
-			${fontWeight ? `font-weight: ${fontWeight}` : ''};
-			${fontStyle ? `font-style: ${fontStyle}` : ''};
-
-			/*
-			 * Child elements (e.g. <Link/>) can use this variable
-			 * to set the thickness of their underline.
-			 *
-			 * The thickness for each font type and weight is defined
-			 * in the underlineThickness object.
-			 */
```
sndrs, line 142: did you mean to remove this comment?

jamesmockett: Yes, as there doesn't seem to be any need to output this comment every time the typography API is called. There's a comment in the source code for this function that explains its purpose which seems sufficient?

sndrs: oh apologies, i missed that, yes agree

sndrs (approved):

PR #1070, opened 2024-01-09
Add theme overrides for ChoiceCard component

oliverabrahams:
## What are you changing?

- Added a Theme Override to the props for the Choice Card component.
- Updated the name of the theme variables to match what is used in [figma](https://www.figma.com/file/wQEC3OHlazYJMk1xdCnVsr/Component-variations-for-CSTI?type=design&node-id=0-1&mode=design&t=9XzDkZ4YcfLoKCBU-0)
- Added a dark mode option to the choice card component

Current way of providing a theme:
```
<ThemeProvider theme={myTheme}>
  <ChoiceCard>
    Select
  </ChoiceCard>
</ThemeProvider>
```

Updated way of providing a theme. Any variables you supply in the object are optional. You can override one or multiple of the defaults 
```
const myTheme: ChoiceCardTheme = {backgroundSelected: palette.brand[500]}
<ChoiceCard theme={myTheme}>
  Select
</ChoiceCard>
```

## Why?

- WebX need to implement dark mode in DCR for apps. 
- This is a proof of concept for a different way for the theme to be provided to components in source-react-components


libs/@guardian/source-react-components/src/choice-card/ChoiceCard.stories.tsx
```diff
@@ -35,7 +37,38 @@ export default {
 };
 
 const Template: Story<ChoiceCardProps> = (args: ChoiceCardProps) => (
-	<ChoiceCard {...args} />
+	<div style={{ columns: 5 }}>
```
joecowton1, line 40: Are these better as 3 separate stories?

libs/@guardian/source-react-components/src/choice-card/ChoiceCard.stories.tsx
```diff
@@ -35,7 +37,38 @@ export default {
...
+		<ChoiceCard {...args} />
+
+		{/* Can use all available values */}
+		<ThemeProvider
+			theme={{
+				choiceCard: {
+					...choiceCardThemeDefault.choiceCard,
+					backgroundUnSelected: 'green',
```
joecowton1, line 48: could we use Guardian palette colours in the examples?

libs/@guardian/source-react-components/src/choice-card/ChoiceCard.tsx
```diff
@@ -49,6 +55,10 @@ export interface ChoiceCardProps
 	 * The type of input you want
 	 */
 	type?: 'radio' | 'checkbox';
+	/**
+	 * A component level theme to override the colour palette of the button
+	 */
+	theme?: ChoiceCardTheme | 'light' | 'dark';
```
joecowton1, line 61: I think i imagined we were going to treat any 'dark' theme as the just the same as any other theming option - if we're specifying it here then that gives it some kind of special status

libs/@guardian/source-react-components/src/choice-card/ChoiceCard.tsx
```diff
@@ -81,14 +92,50 @@ export const ChoiceCard = ({
...
+	};
 	// prevent the animation firing if a Choice Card has been checked by default
 	const [userChanged, setUserChanged] = useState(false);
 
 	return (
 		<>
 			<input
 				css={(theme: Theme) => [
-					input(theme.choiceCard),
+					input(getCombinedTheme(transformProvidedTheme(theme.choiceCard))),
```
joecowton1, line 138: it looks like we always call these functions together - can we not just put the transform inside getCombined theme?

libs/@guardian/source-react-components/src/choice-card/ChoiceCard.tsx
```diff
@@ -81,14 +92,50 @@ export const ChoiceCard = ({
...
+	): ChoiceCardFullTheme => {
+		if (typeof theme !== 'string' && theme) {
+			return {
+				...choiceCardThemeLight,
+				...providedTheme,
+				...theme,
+			};
+		}
+		if (typeof theme === 'string') {
+			switch (theme) {
```
joecowton1, line 122: see earlier comment - this feels wrong to have a separate switch between light and dark mode. are we not just wanting to give the option of passing a dark mode theme in?

joecowton1 (approved):

libs/@guardian/source-react-components/src/choice-card/theme.ts
```diff
@@ -12,10 +27,40 @@ export const choiceCardThemeDefault = {
...
-		textHover: palette.brand[400],
+		textHover: palette.brand[500],
 		borderHover: palette.brand[500],
 		textError: palette.error[400],
 		borderError: palette.error[400],
 	},
 	...userFeedbackThemeDefault,
 };
+
+export const choiceCardThemeDark = {
```
joecowton1, line 38: see other comments about making dark mode a special case

libs/@guardian/source-react-components/src/choice-card/theme.ts
```diff
@@ -1,6 +1,21 @@
...
+	borderUnSelected?: string;
+	borderSelected?: string;
+	borderHover?: string;
+	borderError?: string;
+	backgroundUnSelected?: string;
+	backgroundHover?: string;
+	backgroundSelected?: string;
+	backgroundTick?: string;
+}
+/** @deprecated use ChoiceCardThemeLight **/
```
jamesmockett, line 18: Should the deprecation warning say something about not using `ThemeProvider` too as we want people to use the `theme` prop rather than simply swapping themes?

```suggestion
/** @deprecated Use `choiceCardThemeLight` and `theme` prop instead of `ThemeProvider` **/
```

libs/@guardian/source-react-components/src/choice-card/ChoiceCard.stories.tsx
oliverabrahams: Moved this to a new story. What do you think?

oliverabrahams: Actually just deleted it

libs/@guardian/source-react-components/src/choice-card/theme.ts
```diff
@@ -12,10 +27,25 @@ export const choiceCardThemeDefault = {
...
-		textHover: palette.brand[400],
+		textHover: palette.brand[500],
 		borderHover: palette.brand[500],
 		textError: palette.error[400],
 		borderError: palette.error[400],
 	},
 	...userFeedbackThemeDefault,
 };
+
+export const choiceCardTheme = {
```
joecowton1, line 38: we could type this fwiw

oliverabrahams: done

libs/@guardian/source-react-components/src/choice-card/ChoiceCard.stories.tsx
oliverabrahams: Do you think 2 stories, One for recommended theming and one for backwards compatibility. Or just one for recommended

oliverabrahams: Screenshot 2024-01-11 at 16 43 29

oliverabrahams: https://635a6fffacd30d393597c1ff-uixqjbdlzi.chromatic.com/?path=/docs/choicecard-theming--docs

joecowton1: Sorry Ollie, actually thinking about it this might get a bit much if we do this for each component. Can we just have an additional story within the ChoiceCard stories that demonstrates an alternative theme?

oliverabrahams: :'(

oliverabrahams: https://635a6fffacd30d393597c1ff-qtrtddfvkq.chromatic.com/?path=/docs/choicecard--docs

libs/@guardian/source-react-components/src/choice-card/theme.ts
```diff
@@ -1,6 +1,25 @@
...
+	borderHover?: string;
+	borderError?: string;
+	backgroundUnselected?: string;
+	backgroundHover?: string;
+	backgroundSelected?: string;
+	backgroundTick?: string;
+}
+
+export type ChoiceCardFullTheme = {
+	[P in keyof ChoiceCardTheme]-?: ChoiceCardTheme[P];
```
sndrs, line 20: what is this doing? can't quite get mu head around it

libs/@guardian/source-react-components/src/choice-card/styles.ts
```diff
@@ -61,9 +61,7 @@ export const gridColumns: { [key in ChoiceCardColumns]: SerializedStyles } = {
 	5: gridColumnsStyle(5),
 };
 
-export const input = (
-	choiceCard = choiceCardThemeDefault.choiceCard,
-): SerializedStyles => css`
+export const input = (choiceCard = choiceCardTheme): SerializedStyles => css`
```
sndrs, line 64:
```suggestion
export const input = (theme = choiceCardTheme): SerializedStyles => css`
```

libs/@guardian/source-react-components/src/choice-card/theme.ts
oliverabrahams: This just provides a type with all the new properties without the optional `?` which is used for the function which returns the theme to the functions in styles.ts

sndrs (approved):

PR #1071, opened 2024-01-09
Identity Auth | Change `isSignedIn` check logic

coldlink:
## What are you changing?

- Updates the `isSignedIn` check logic to use the `GU_SO` cookie correctly
- Clear the `GU_U` cookie if we fail to get tokens, as it's likely invalidly set

## Why?

- We discovered an issue in the way that the `GU_SO` cookie works, meaning that a user could sign out, then immediately sign in again as a new user, but the tokens for the old user had not been cleared yet.
  - The problem and solution have been outlined in detail within this PR: https://github.com/guardian/gateway/pull/2529
  - And the documentation was updated in this PR: https://github.com/guardian/gateway/pull/2530
- Regarding the `GU_U` cookie, we should clear this cookie if we fail to get tokens, as otherwise the library will keep trying to get tokens, causing unnecessary API calls, and could if this affected multiple users at once cause our API limits to be reached on the token endpoint

### DIFF
--- a/libs/@guardian/identity-auth/src/identityAuth.ts
+++ b/libs/@guardian/identity-auth/src/identityAuth.ts
@@ -1,4 +1,4 @@
-import { getCookie } from '@guardian/libs';
+import { getCookie, removeCookie } from '@guardian/libs';
 import type {
 	IdentityAuthOptions,
 	IdentityAuthState,
@@ -153,6 +153,10 @@ export class IdentityAuth<
 						idToken: tokens.idToken,
 						isAuthenticated: true,
 					};
+				} else {
+					// if we weren't able to get the tokens, despite having a GU_U cookie, the user is not signed in
+					// we should clear the GU_U cookie, as it is likely invalid
+					removeCookie({ name: 'GU_U' });
 				}
 			}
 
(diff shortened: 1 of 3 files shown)


libs/@guardian/identity-auth/src/identityAuth.ts
```diff
@@ -94,23 +94,31 @@ export class IdentityAuth<
...
+			// This follows the flowchart from https://github.com/guardian/gateway/blob/main/docs/okta/web-apps-integration-guide.md#how-to-know-if-a-reader-is-signed-in to determine if the user is signed in
+			// first check if the user has valid access and id tokens
 			const authState = this.authStateManager.getAuthState();
+			// if they do we need to check if a user has recently signed out
 			if (authState.isAuthenticated) {
 				// validate the id token and access token to make sure auth state is still valid
 				await this.token.verifyTokens(authState.idToken, authState.accessToken);
 
+				// if a user has a GU_SO cookie, they might have recently signed out
+				const guSoCookie = parseInt(
```
raphaelkabo, line 106: If the `parseInt` call fails, this gets caught all the way at the end - is that acceptable?

libs/@guardian/identity-auth/src/identityAuth.ts
```diff
@@ -94,23 +94,31 @@ export class IdentityAuth<
 	 */
```
raphaelkabo, line 94: You may want to update the docblock here - on line 87 it reads `If the user has a GU_SO cookie, they have recently signed out, so we should clear their tokens (side effect)`, which is not exactly true any more.

coldlink: very good spot!

will update the comment

libs/@guardian/identity-auth/src/identityAuth.ts
raphaelkabo: Perfect!

raphaelkabo (approved):
Looks super!

PR #1076, opened 2024-01-11
Update `InlineError` warning icon

jamesmockett:
## What are you changing?

- Updates triangular warning icon with circular version.

## Why?

- The new rounded icon is now in keeping with the icon set and it also future-proofs us for using these icons as navigation badges.

### Before

Screenshot 2024-01-11 at 17 29 41

### After

Screenshot 2024-01-11 at 17 29 36


mxdvl (approved):
🔴

PR #1085, opened 2024-01-16
Update to `theme` prop proposal

sndrs:
## What are you changing?

- create one `choiceCardTheme` type
- renames some of the theme munging stuff
- requires a theme to be passed the styles functions (rather than defaulting - let's default in one place)
- adds types the to story to clean up the red squiggles

### DIFF
--- a/libs/@guardian/source-react-components/src/choice-card/ChoiceCard.stories.tsx
+++ b/libs/@guardian/source-react-components/src/choice-card/ChoiceCard.stories.tsx
@@ -1,5 +1,4 @@
-import { palette } from '@guardian/source-foundations';
-import type { Story } from '@storybook/react';
+import type { Meta, StoryFn } from '@storybook/react';
 import { SvgCamera } from '../../vendor/icons/SvgCamera';
 import { ChoiceCard } from './ChoiceCard';
 import type { ChoiceCardProps } from './ChoiceCard';
@@ -18,7 +17,7 @@ const choiceCardThemeDark = {
 	backgroundSelected: palette.neutral[100],
 	backgroundTick: palette.brand[500],
 };
-export default {
+const meta: Meta<typeof ChoiceCard> = {
 	title: 'ChoiceCard',
 	component: ChoiceCard,
 	args: {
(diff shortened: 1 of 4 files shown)


libs/@guardian/source-react-components/src/choice-card/ChoiceCard.stories.tsx
```diff
@@ -82,7 +69,10 @@ IconDefaultTheme.args = {
...
 
-export const DarkTheme = Template.bind({});
-DarkTheme.args = {
-	theme: choiceCardThemeDark,
+export const CustomTheme: StoryFn<ChoiceCardProps> = Template.bind({});
+CustomTheme.args = {
+	theme: {
+		backgroundUnselected: 'black',
+		backgroundSelected: 'hotpink',
+	},
```
oliverabrahams, line 77: is the `theme: choiceCardDark` better if we consider this as part of the documentation for this component. It is the "recommended" way of theming

libs/@guardian/source-react-components/src/choice-card/ChoiceCard.tsx
```diff
@@ -92,30 +102,25 @@ export const ChoiceCard = ({
...
+	const transformOldProviderTheme = (
+		providerTheme: Theme['choiceCard'],
+	): Partial<ChoiceCardTheme> => ({
+		textUnselected: providerTheme?.textLabel,
+		textSelected: providerTheme?.textChecked,
+		borderUnselected: providerTheme?.border,
+		borderSelected: providerTheme?.borderChecked,
+		backgroundSelected: providerTheme?.backgroundChecked,
+		...providerTheme,
+	});
```
oliverabrahams, line 115:
```suggestion
	const transformOldProviderTheme = (
		providerTheme: Theme['choiceCard'],
	): Partial<ChoiceCardTheme> => {
		const transformedTheme: Partial<ChoiceCardTheme> = {};

		if (providerTheme?.textLabel) {
			transformedTheme.textUnselected = providerTheme.textLabel;
		}
		if (providerTheme?.textChecked) {
			transformedTheme.textSelected = providerTheme.textChecked;
		}
		if (providerTheme?.border) {
			transformedTheme.borderUnselected = providerTheme.border;
		}
		if (providerTheme?.borderChecked) {
			transformedTheme.borderSelected = providerTheme.borderChecked;
		}
		if (providerTheme?.backgroundChecked) {
			transformedTheme.backgroundSelected = providerTheme.backgroundChecked;
		}
		return { ...transformedTheme, ...providerTheme };
	};
```

oliverabrahams: This stops these properties from being overwritten by undefined

libs/@guardian/source-react-components/src/choice-card/ChoiceCard.stories.tsx
oliverabrahams: That makes sense. Happy to go with that. I had something similar before, we could change the colours to come from the palette because that may be more how we want the prop used.
```suggestion
CustomTheme.args = {
	theme: {
		backgroundUnselected: palette.neutral[0],
		backgroundSelected: palette.lifestyle[500],
	},
```

oliverabrahams (approved):
LGTM

libs/@guardian/source-react-components/src/choice-card/ChoiceCard.stories.tsx
sndrs: yeah i think we do want that, but part of me felt that because right now you _can_ do the 'wrong' thing, we should have a story that does that, then have it break when we fix it. maybe that's the wrong approach.

but maybe we shoudl fix the types in foundations so you can't?

PR #1090, opened 2024-01-18
ci: Use first-party GitHub Action

akash1810:
## What are you changing?
Replaces a third-party GitHub Action with a first-party[^1] one.

[^1]: See also https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions.


joecowton1 (approved):
Thanks Akash!

PR #1091, opened 2024-01-19
Fix Storybook type annotations

jamesmockett:
## What are you changing?

- Updates type annotations in Source Storybook stories.

## Why?

- All of the stories have a number of type errors. Whilst these (mostly) don't prevent the stories from running or rendering in Chromatic, the errors create a lot of visual noise in the code editor and can cause confusion when working on them.


joecowton1 (approved):
Nice one James, that's a lot of type annotations! Thanks for sorting this.

PR #1092, opened 2024-01-19
Add theme prop to checkbox component and deprecate old themes

oliverabrahams:
## What are you changing?

- Added new theme prop to the checkbox component

I tried to follow the naming of the component listed in [Figma](https://www.figma.com/file/wQEC3OHlazYJMk1xdCnVsr/Component-variations-for-CSTI?type=design&node-id=0-1&mode=design&t=DdpYX4PSNszh3Kpy-0) I struggled to match the all names with properties in the component. there is `fill selected` and `fill unselected` which i changed to `background` and `tick` which made me wonder what the value was in changing these names to what is in Figma. 

#### Pros:

- Clarity for Developers: Using descriptive names makes it easier for developers to understand the function of each option.
- Design Consistency: Aligning names with those used in design documents makes it easier for developers to translate the designs into code

#### Cons:

- Risk and Complexity: Altering the object structure to match Figma's naming introduces potential risks and adds complexity, especially in supporting both the new and old structures.

### Proposed Options for Property Naming:

1. Developer-Friendly Names: Opt for names that are intuitive and easily understood by developers and consumers of the component.
2. Figma Names: Use names that match those used by our designers.I wonder about the standardization of these names across designs / how often will they be changed.
3. Keeping Current Names: Retain the existing names for continuity and add any new properties that are missing.

In this PR I have gone for option 1 (developer friendly names) I am happy to do whatever the team thinks is best. Just putting down some thoughts.


libs/@guardian/source-react-components/src/checkbox/Checkbox.tsx
```diff
@@ -90,32 +98,91 @@ export const Checkbox = ({
...
+		if (providerTheme?.backgroundChecked) {
+			transformedTheme.backgroundTick = providerTheme.backgroundChecked;
+		}
+		if (providerTheme?.borderChecked) {
+			transformedTheme.borderSelected = providerTheme.borderChecked;
+		}
+		if (providerTheme?.border) {
+			transformedTheme.borderUnselected = providerTheme.border;
+		}
+		console.log({ ...transformedTheme, ...providerTheme });
```
jamesmockett, line 115: A `console.log` has sneaked in here.

libs/@guardian/source-react-components/src/checkbox/theme.ts
```diff
@@ -34,3 +47,27 @@ export const checkboxThemeBrand = {
...
+
+export const checkboxTheme: CheckboxTheme = {
+	borderUnselected: palette.neutral[46],
+	borderHover: palette.brand[500],
+	borderSelected: palette.brand[500],
+	borderError: palette.error[400],
+	backgroundTick: palette.brand[500],
+	backgroundUnselected: 'transparent',
+	textLabel: palette.neutral[7],
+	textLabelSupporting: palette.neutral[46],
```
jamesmockett, line 59: Similarly to above, the Radio button theme is using `textSupporting` to match Figma.

PR #1097, opened 2024-01-22
Fix file input error border width

joecowton1:
## What are you changing?

- fixing file input border width

## Why?

- it was too wide

## Before
Screenshot 2024-01-22 at 14 46 07

## After
Screenshot 2024-01-22 at 14 45 59


.changeset/forty-sheep-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/source-react-components-development-kitchen': patch
```
jamesmockett, line 2: As this is a visual change that _could_ affect layout it feels like it should be a minor change at least?

jamesmockett (approved):

.changeset/forty-sheep-refuse.md
joecowton1: its a fix not a change, no?

sndrs (approved):

.changeset/forty-sheep-refuse.md
sndrs: agree with james, you can see the impact in the diff 
image

PR #1099, opened 2024-01-22
Fix numeric storage `expires`

tomrf1:
## What are you changing?
Currently if you give `set` a numeric expiry then the item is never expired.

For consistency I've also changed `set` to always add the expiry as a datetime string. But we need `get` to handle numeric values to fix any browsers that currently have that in their local storage

### DIFF
--- a/.changeset/funny-jars-pump.md
+++ b/.changeset/funny-jars-pump.md
@@ -1,0 +1,5 @@
+---
+'@guardian/libs': patch
+---
+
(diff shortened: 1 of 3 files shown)


mxdvl (approved):
I’m wondering if we should also coerce the `expires` to a `Date` in the `set`, so that it is saved consistently

mxdvl (commented):
We're actually missing a test for passing a date `string` directly

PR #1101, opened 2024-01-22
Reduce `xsmall` button text size

jamesmockett:
## What are you changing?

- Reduces text size of `xsmall` button to `14px` (`textSans.xsmall`).
- Removes extra spacing that was being added to outside of `xsmall` buttons if they included a left-aligned icon. 

### Before

Screenshot 2024-01-22 at 17 07 29

### After

Screenshot 2024-01-22 at 17 05 31

### DIFF
--- a/.changeset/nine-owls-pay.md
+++ b/.changeset/nine-owls-pay.md
@@ -1,0 +1,5 @@
+---
+'@guardian/source-react-components': minor
+---
+
+Reduces text size of `xsmall` buttons to `14px`.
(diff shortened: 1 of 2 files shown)


mxdvl (dismissed):
cute as can be

sndrs (approved):

mxdvl:
How was #1097 `major` but this `minor`?

jamesmockett:
> How was #1097 `major` but this `minor`?

#1097 is a tiny change really, but the thinking was that the change in border width causes layout shift and could create larger diffs in Chromatic due to everything shifting by a few pixels. The changes to the button are internal to that component so shouldn't affect the rest of the page layout or result in large diffs. We've debated this and had differing opinions, but we definitely need to be more consistent.

PR #1106, opened 2024-01-23
Change scope of design tokens package to `@guardian`

sndrs:
## What are you changing?

- converting `@csnx/design-tokens` to `@guardian/design-tokens`

## Why?

- changesets ignore `@csnx` packages because they're internal (repo) packages. We want this to be treated as a full guardian package, just not yet published


libs/@guardian/design-tokens/package.json
```diff
@@ -1,9 +1,9 @@
 {
-	"name": "@csnx/design-tokens",
-	"version": "0.0.0",
+	"name": "@guardian/design-tokens",
```
sndrs, line 2: should this be `@guardian/source-design-tokens`?

sndrs: can fix this later

jamesmockett (approved):
👍

PR #1107, opened 2024-01-24
Create `plugin-ts` for cobalt

sndrs:
## What are you changing?

- adds a plugin for cobalt that uses the TS compiler to create js tokens as a TS `const`

## Why?

the default plugin uses a set of complex functions to just return basic types (`string` etc). 

this restores the previous `const` definition we had, which allows us to narrow types to exact tokens.


mxdvl (approved):
thanks muchly!

sndrs:
> thanks muchly!

apologies, i've change this to create a new package and changed the base to main.

helps keep this and #1106 cleaner

libs/@csnx/design-tokens/tokens/index.js
```diff
@@ -143,7 +143,6 @@ export const tokens = {
 			100: '#3E3323',
 			200: '#574835',
 			300: '#6B5840',
-			/** Deprecated: Use colour.culture.400 */
```
sndrs, line 146: this isn't good :(

mxdvl: Should this be `/** @deprecated use colour.culture[400] */`

libs/@guardian/cobalt-plugin-ts/index.mjs
```diff
@@ -0,0 +1,92 @@
...
+			// now we create a ts program to create the JS and declaration file
+			// contents. the program would write to disk by default, so we
+			// override that behaviour to capture the output in memory instead.
+
+			// this is where we'll store the output for the JS and declaration
+			// files
+			const output = {};
+
+			const program = ts.createProgram(
+				['fakefile.ts'],
```
mxdvl, line 45: Would extracting this to a const help with maintainability? Maybe even assign it to a `Symbol`

libs/@guardian/source-foundations/src/utils/convert-value.ts
```diff
@@ -1,5 +1,7 @@
 export const rootPixelFontSize = 16;
 export const pxToRem = (px: number): number => px / rootPixelFontSize;
 export const pxStringToNumber = (px: string): number => Number(px.slice(0, -2));
-export const fontArrayToString = (fonts: string[]): string =>
+export const fontArrayToString = (
+	fonts: string[] | readonly string[],
+): string =>
```
mxdvl, line 6: A mutable string is always acceptable where a readonly one can be passed, so we only need to make the function signature accept immutable ones:

```suggestion
export const fontArrayToString = (
	fonts: readonly string[],
): string =>
```

[TS Playground](https://www.typescriptlang.org/play?target=99#code/MYewdgziA2CmB0w4EMBOAKAlAbgFC9EgBcACAWwFcjkAjOEgXhIG0BGAGhICZOBmTgCycArJwBsAXTyEIpAJZlK1OrEYsO3PoJHiJJZBBIyieAuFkkADqhDBYEQ03RpUyAJ4AuEqljIAJuDQbiRgFGQ0sKjMEpiMAHz6qK5u8GTIluih4ZHxIWERqPBEIADKRKhyYADm6FyYmPgyMAjQIDXWtvYQ6Eq0cPVmkM3wre02dg7oCr0q9UA)

sndrs: that doesn't work when it's a readonly prop:

image

[TS Playgorund](https://www.typescriptlang.org/play?target=99#code/MYewdgziA2CmB0w4EMBOAKAlAbgFC9EgBcACADxIF4SBvXEhkgTwC4SBtAcgAsBLTgLq4AviWQQShCEXwAzAK5hgRXuBIAvdMjbTUvMAHN2AzLRLD8msvCaYgA)

libs/@guardian/source-foundations/src/typography/data.ts
```diff
@@ -14,33 +14,33 @@ import type { AvailableFontWeights, ItalicsFontWeights } from './types';
...
 	textSans: {
-		xxsmall: pxStringToNumber(tokens.typography.textSans.xxSmall.fontSize!),
-		xsmall: pxStringToNumber(tokens.typography.textSans.xSmall.fontSize!),
-		small: pxStringToNumber(tokens.typography.textSans.small.fontSize!),
-		medium: pxStringToNumber(tokens.typography.textSans.medium.fontSize!),
-		large: pxStringToNumber(tokens.typography.textSans.large.fontSize!),
-		xlarge: pxStringToNumber(tokens.typography.textSans.xLarge.fontSize!),
-		xxlarge: pxStringToNumber(tokens.typography.textSans.xxLarge.fontSize!),
-		xxxlarge: pxStringToNumber(tokens.typography.textSans.xxxLarge.fontSize!),
+		xxsmall: pxStringToNumber(tokens.typography.textSans.xxSmall.fontSize),
```
jamesmockett, line 17: It's nice that these no longer require a non-null assertion ❤️

jamesmockett (approved):
This is great and opens the door to refining the output so we don't require as many transforms in Source Foundations 👍

libs/@guardian/source-foundations/src/utils/convert-value.ts
mxdvl: So my point is that the method `fontArrayToString` never relies on the fact that the `fonts` array it receives is mutable, so it’s safe to update its signature to only consider a `readonly string[]`. This is because all mutable arrays can be provided where an immutable one is required, but not the other way around.

[TS Playground](https://www.typescriptlang.org/play?target=99#code/MYewdgziA2CmB0w4EMBOAKAlAbgFC9EgBcACADxIF4SBvXEhkgTwC4SBtAcgAsBLTgLq4AviWQQShCEXxTSZAExVa9Rqw6cARk1iCANCLy4AZgFcwwIr3AkAXumRtUsZABNw0JiWmpeYAObsApi0JML49mTwTJi4kQrRmEA)

PR #1122, opened 2024-01-29
Spike a single `@guardian/source` package

### LINKED ISSUE #1118: Spike a single `@guardian/source` package
Managing simultaneous releases to the various source packages is complex, and for consumers, having to keep them in sync is time consuming with little benefit.

the only reason they are separate now is because we didn't know about `peerDependenciesMeta` when we released v4, and didn't want foundations to have a peerDep on react; and because `targets` was not well supported.

neither of those things should be an issue now.

### `@guardian/source`
equivalent to `@guardian/source-foundations`, e.g.

```js
import { transitions } from "@guardian/source"
```

### `@guardian/source/react-components`
equivalent to `@guardian/source-react-components`, e.g.

```js
import { Checkbox } from "@guardian/source/react-components"
```

### `@guardian/source/react-components/development-kitchen`
equivalent to `@guardian/react-components-development-kitchen`, e.g.

```js
import { NumericInput } from "@guardian/source/react-components/development-kitchen"
```

sndrs:
## What are you changing?

- updates `@csnx/npm-package` to accept entries as objects
- create a single `@guardian/source` package

## Why?

the current division of 3 packages was based on things which no longer apply: 
- poor support for `exports` in package.json (at the time)
- not realising that `peerDependenciesMeta` meant you didn't have to install peer deps of a single package if you weren't using them (e.g. react)

This is a spike to see what a single package would feel like.

## How does it compare?

This package makes use of [package exports](https://webpack.js.org/guides/package-exports/#support) to keep the current family setup:

| Current package                                         | Single package export                                           |
| --- | --- |
| `@guardian/source-foundations` | `@guardian/source` |
| `@guardian/source-react-components` | `@guardian/source/react` |
| `@guardian/source-react-components-development-kitchen` | `@guardian/source/react/development-kitchen` |

closes #1118


joecowton1 (approved):
Looks great Alex.

If we decide to go with this in the long run would we leave it like this permanently or look to change the structure?

sndrs:
> If we decide to go with this in the long run would we leave it like this permanently or look to change the structure?

I'd imagine the exports structure would stay (if we all like it) but the code structure would coalesce into the main dir. 

the setup in this PR allows us to keep publishing the current packages as normal. If we stop doing that and go with this, I think it would make sense to move the code

PR #1123, opened 2024-01-30
Make media queries’ `max-width` much closer to the breakpoint value

mxdvl:
## What are you changing?

- Make media queries’ max-width much closer to the value, just 0.001px shy.
- Live demonstration: https://codepen.io/mxdvl/pen/XWGZmMm

## Why?

- As media queries can report fractional values, they fall between entire pixels. For example, `479.5px` is matching neither `max-width: 479px` nor `min-width: 480px`
- https://github.com/guardian/dotcom-rendering/issues/10195
- The media range syntax would be more expressive, but [support is sparse at ~80%](https://caniuse.com/css-media-range-syntax)

### DIFF
--- a/.changeset/cyan-trainers-brush.md
+++ b/.changeset/cyan-trainers-brush.md
@@ -1,0 +1,10 @@
+---
+'@guardian/source-foundations': patch
+---
+
+Make media queries `max-width` much closer to the breakpoint value.
+
+As media queries can report fractional values, they can currently fall between entire pixels.
+For example, `479.5px` is matching neither `max-width: 479px` nor `min-width: 480px`
+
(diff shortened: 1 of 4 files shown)


mxdvl:
> Could this be solved by using rem values instead ?

No, it would not, for two reasons:
1. `rem` values in media queries as [simply multiplied by `16`](https://smth.uk/rem-based-media-queries-are-weird/) regardless of the `html`’s `font-size`.
2. fractional values can still fall between entire values. e.g. `29.6rem` would match neither `max-width: 29.5rem` nor `min-width: 29.75rem`

sndrs (approved):

PR #1128, opened 2024-01-30
Move CMP to `@guardian/libs`

vlkerag:
## What are you changing?

Adds the exports of `@guardian/consent-management-platform` to `@guardian/libs`.

_The original code is marked as being moved here in https://github.com/guardian/consent-management-platform/pull/960._

## Why?

We want to add consent-based purpose checks to exports of `@guardian/libs` that write to user's devices (cookies, storage etc).

Because the CMP already depends on `libs`, consuming the CMP from `libs` would create [a circular dependency](https://dev.to/successgilli/circular-dependency-in-nodejs-and-nestjs-3e1d), which makes it impossible to manage updates etc.

Therefore, we need to co-locate this code.

Refs [#1326](https://github.com/guardian/csnx/issues/1326).

> [!NOTE]
>
> The following changes to the CMP code were required:
>
> - Remove extensions from CMP import specifiers
> - Imports (including mocks) from `@guardian/libs` were updated to local paths, e.g. 
>   ```ts
>   import { log } from '../logger/logger';
>   
>   jest.mock("../storage/storage", () => ({
>     // ...
>   }));
>   ```
> - Types renamed (including exports):
>   - `Callback` → `OnConsentChangeCallback`
>   - `Framework` → `ConsentFramework`


libs/@guardian/libs/cypress/support/e2e.js
```diff
@@ -14,4 +14,4 @@
 // ***********************************************************
 
 // Import commands.js using ES2015 syntax:
-import './commands';
+import '../e2e/consent-management-platform/support/commands';
```
oliverabrahams, line 17: Do we want to have e2e.js and comands.js in separate directories

oliverabrahams (approved):
Looks good to me!

akinsola-guardian (approved):
Thanks for this. This looks good.

libs/@guardian/libs/src/index.ts
```diff
@@ -58,3 +58,21 @@ export { storage } from './storage/storage';
 
 export type { Switches } from './switches/@types/Switches';
 export { getSwitches } from './switches/getSwitches';
+
+export type {
+	Callback,
```
sndrs, line 63: i think this name is a bit too generic for a library with many other exports - how about something like `OnConsentChangeCallback`?

libs/@guardian/libs/src/index.ts
```diff
@@ -58,3 +58,21 @@ export { storage } from './storage/storage';
...
+	CMP,
+	ConsentState,
+	Framework,
```
sndrs, line 66: this too is too generic - how about `ConsentFramework`?

sndrs (changes requested):
we need to add a lint to https://github.com/guardian/consent-management-platform for the merge commit.

i'm on it, just hitting the red button to hold off merging till then

libs/@guardian/libs/src/index.ts
```diff
@@ -58,3 +58,21 @@ export { storage } from './storage/storage';
 
 export type { Switches } from './switches/@types/Switches';
 export { getSwitches } from './switches/getSwitches';
+
+export type {
+	Callback as OnConsentChangeCallback,
```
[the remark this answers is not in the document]
sndrs: yeah agreed - but given you'll need to import from `libs` now, should that means a major version of `libs`?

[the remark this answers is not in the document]
mxdvl: Oh yes, true! Carry on…

sndrs (approved):

sndrs:
will merge on monday morning, alongside https://github.com/guardian/consent-management-platform/pull/960

PR #1129, opened 2024-01-30
Enable changeset `bumpVersionsWithWorkspaceProtocolOnly` option again

jamesmockett:
## What are you changing?

- Enable changesets `bumpVersionsWithWorkspaceProtocolOnly` option, reverting change in #1120

## Why?

- Whilst enabling this option gives us the behaviour we want in terms of bumping dependencies and peer dependencies it does not update the lock file at the same file causing the build to fail when merged.
- We may be able to update the lock file with an action that's triggered by a package release PR, but in the meantime it seems safest to revert to the existing behaviour that we're familiar with.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -7,7 +7,7 @@
 	"access": "public",
 	"baseBranch": "main",
 	"updateInternalDependencies": "patch",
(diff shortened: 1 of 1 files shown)


oliverabrahams (approved):
LGTM

PR #1130, opened 2024-01-30
Add `dev` tasks to packages with unit tests

sndrs:
## What are you changing?

- adds `dev` tasks where meaningful
- adds a `make dev` task that can run from the root

## Why?

- should make it easier to get instant feedback on changes


oliverabrahams (approved):
Looks good!

libs/@guardian/libs/project.json
```diff
@@ -34,7 +34,7 @@
 			"outputs": ["{workspaceRoot}/coverage/libs/@guardian/libs"],
 			"options": {
 				"jestConfig": "libs/@guardian/libs/jest.config.ts",
-				"passWithNoTests": true
+				"passWithNoTests": false
```
jamesmockett, line 37: Assume this is intentional as it's `true` for other projects? (Clearly there _are_ tests for libs anyway.)

sndrs: doh! that slipped through from #1131

jamesmockett: If #1131 is going in next then there's probably no point changing this back 👍

oliverabrahams (approved):
LGTM

PR #1131, opened 2024-01-30
Require unit tests where jest is configured

sndrs:
## What are you changing?

- stops projects having no tests

## Why?

- if we've configured jest, we should us it


oliverabrahams (approved):
LGTM

PR #1133, opened 2024-01-30
Update README to show new location for husky config

oliverabrahams:
## What are you changing?

- With the updated version of husky in csnx the ~/.huskyrc file needs to be moved to ./config/husky/init.sh 
- In new version the below lines can be removed 

```
#!/bin/sh

. "$(dirname "$0")/_/husky.sh"
```
## Why?

- New version of husky https://github.com/typicode/husky/releases/tag/v9.0.1

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,4 +248,4 @@ export NVM_DIR="$HOME/.nvm"
 [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh" && nvm use
 ```
 
-See https://typicode.github.io/husky/#/?id=command-not-found for more info.
+See https://typicode.github.io/husky/how-to.html for more info.
--- a/README.md
+++ b/README.md
@@ -210,12 +210,15 @@ This will create a "changeset": a `.md` file containing the release information.
 
 ### Unable to commit
 
-If you get a `command not found` error or a message saying you're using the wrong version of Node when commiting using a GUI (VSCode, GitHub desktop etc), add a `~/.huskyrc` file and load your Node version manager there.
+> [!IMPORTANT]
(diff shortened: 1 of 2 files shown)


sndrs:
do you also want to update the `prepare` npm-script to `"prepare": "husky"` to clear up the deprecation notice when installing node modules?

sndrs (changes requested):
do you also want to update the prepare npm-script to "prepare": "husky" to clear up the deprecation notice when installing node modules?

README.md
```diff
@@ -212,39 +212,42 @@ This will create a "changeset": a `.md` file containing the release information.
 
 ### Unable to commit
 
-If you get a `command not found` error or a message saying you're using the wrong version of Node when commiting using a GUI (VSCode, GitHub desktop etc), add a `~/.huskyrc` file and load your Node version manager there.
+> [!IMPORTANT]
+> The ~/.huskyrc file has moved to ~/.config/husky/init.sh
+
+If you get a `command not found` error or a message saying you're using the wrong version of Node when commiting using a GUI (VSCode, GitHub desktop etc), add a `~/.config/husky/init.sh` file and load your Node version manager there.
```
sndrs, line 218: this is really nice. I wonder if it makes sense to put the note about the file having moved after we first mention it?

```suggestion
If you get a `command not found` error or a message saying you're using the wrong version of Node when commiting using a GUI (VSCode, GitHub desktop etc), add a `~/.config/husky/init.sh` file and load your Node version manager there.

> [!NOTE]
> This used be located in `~/.huskyrc`. If you set that up before, you will need to recreate it at `~/.config/husky/init.sh`.
> ```sh
> mkdir -p ~/.config/husky && cp ~/.huskyrc $_/init.sh
> ```
```

sndrs (approved):

PR #1134, opened 2024-01-31
use => in storage for functions

vlkerag:
## What are you changing?

- storage uses `=>` notation

## Why?

This change eliminates the following when moving the cmp into libs:

```ts
// libs/@guardian/libs/src/consent-management-platform/vendorDataManager.test.ts
vendorStorageIds.permutive.localStorage.forEach((name) => {
	expect(storage.local.remove).toHaveBeenCalledWith(name);
});
```

leads to the following type error that will be eliminated by this PR:

```
@typescript-eslint/no-unsafe-return
         88:11  error  Avoid referencing unbound methods which may cause unintentional scoping of `this`.
       If your function does not access `this`, you can annotate it with `this: void`, or consider using an arrow function instead 
```

### DIFF
--- a/.changeset/unlucky-toes-argue.md
+++ b/.changeset/unlucky-toes-argue.md
@@ -1,0 +1,5 @@
+---
+"@guardian/libs": patch
+---
+
+use `=>` to always bind storage methods to their parent scope
--- a/.changeset/unlucky-toes-argue.md
+++ b/.changeset/unlucky-toes-argue.md
@@ -1,5 +1,5 @@
 ---
-"@guardian/libs": patch
+'@guardian/libs': patch
 ---
 
 use `=>` to always bind storage methods to their parent scope
--- a/libs/@guardian/libs/src/storage/storage.ts
+++ b/libs/@guardian/libs/src/storage/storage.ts
@@ -20,9 +20,7 @@ class StorageFactory {
 	/**
 	 * Check whether storage is available.
 	 */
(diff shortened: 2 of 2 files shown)


mxdvl:
I think the ESLint rule that it breaks is [`unbound-method`](https://typescript-eslint.io/rules/unbound-method/) rather that `no-unsafe-return`. 

Should we consider whether to add this to `@guardian/typescript-eslint` instead?

sndrs:
> I think the ESLint rule that it breaks is [`unbound-method`](https://typescript-eslint.io/rules/unbound-method/) rather that `no-unsafe-return`.
> 
> Should we consider whether to add this to `@guardian/typescript-eslint` instead?

In theory, we should be:

https://github.com/guardian/csnx/blob/afdcd6003c5ec13ce45ed0f6362693f0f1d10845/libs/%40guardian/eslint-config-typescript/index.js#L13

that config name has changed https://typescript-eslint.io/linting/configs/#recommended-configurations but [it should still apply](https://typescript-eslint.io/blog/announcing-typescript-eslint-v6/#reworked-configuration-names:~:text=recommended%2Drequiring%2Dtype%2Dchecking%20is%20now%20an%20alias%20for%20recommended%2Dtype%2Dchecked.%20The%20alias%20will%20be%20removed%20in%20a%20future%20major%20version.)

sndrs (approved):

PR #1147, opened 2024-02-01
Add `theme` prop support to Radio buttons

jamesmockett:
## What are you changing?

- Adds support for `theme` prop and updated theme format, deprecating use of existing themes and `ThemeProvider`.
- Theming has been added to the `Radio` and `RadioGroup` components.
- The `transformOldProviderTheme` method that translates existing themes to the new format has been moved to the `theme` module so that it can be used by both components without having to duplicate the code.

## Additional information

Brand themed Radio groups currently have an incorrectly coloured label as the `Legend` component has not yet been updated with `theme` prop support.


oliverabrahams (approved):
LGTM

PR #1155, opened 2024-02-05
Add `theme` prop to Icons

oliverabrahams:
## What are you changing?

- Added new theming to Icons

## Why?

-


mxdvl (commented):
Am I correct in thinking that only monochrome icons can be themed?

oliverabrahams:
> Am I correct in thinking that only monochrome icons can be themed?

Yes, I think that makes sense. What do you think?

mxdvl:
> > Am I correct in thinking that only monochrome icons can be themed?
> 
> Yes, I think that makes sense. What do you think?

I think that enabling polychromatic icons to be styled is going to be a much larger endeavour that is best tackled only if consumers demand it.

libs/@guardian/source-react-components/src/icons/Icons.stories.tsx
```diff
@@ -327,19 +331,6 @@ MediumIconsBrandTheme.parameters = {
 		default: 'brandBackground.primary',
 	},
 };
-MediumIconsBrandTheme.decorators = [
```
jamesmockett, line 330: Glad to see these style overrides removed and replaced with a supported method so we're not encouraging people to do the wrong thing 🙈

jamesmockett (approved):
Nice work 🚀

PR #1163, opened 2024-02-06
Add generic theme merging function

jamesmockett:
## What are you changing?

- Adds generic `mergeThemes` function to merge values from new default themes, deprecated `ThemeProvider` themes and theme overrides, optionally passing `ThemeProvider` themes through a transform function to map old token names to new ones.

## Why?

- This logic is duplicated in every component now to support the new `theme` prop alongside the now deprecated `ThemeProvider` themes. Having it in a central place will make removal of the deprecated themes and transform functions simpler.


oliverabrahams (approved):
Funny how this is the same number of lines. It is much better with the tests and the generic function 💯

PR #1177, opened 2024-02-07
Add `theme` prop support to Text Area 

jamesmockett:
## What are you changing?

- Adds support for theming via `theme` prop to `TextArea` component.
- This component does not have any existing support for theming via `ThemeProvider`.

### DIFF
--- a/libs/@guardian/source-react-components/src/text-area/TextArea.tsx
+++ b/libs/@guardian/source-react-components/src/text-area/TextArea.tsx
@@ -15,6 +15,7 @@ import {
 	widthFluid,
 } from './styles';
 import { InputSize } from '../@types/InputSize';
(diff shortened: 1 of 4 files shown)


libs/@guardian/source-react-components/src/text-area/TextArea.tsx
```diff
@@ -98,27 +104,37 @@ export const TextArea = ({
 
 		return undefined;
 	};
+	const mergedTheme = { ...themeTextArea, ...theme };
```
jamesmockett, line 107: This component doesn't have any existing support for `ThemeProvider` theming so there's no need to use the shared `mergeThemes` method here.

oliverabrahams: Nice! yea this is what i need to do with the Icons also

PR #1178, opened 2024-02-07
Add canary release workflow 🐥

sndrs:
## What are you changing?

Adds a canary release option to PRs.

## Why?

So we can test packages in external consumer codebases (@guardian/transparency-consent do this a lot, so it's needed for #1128).

## How does it work?

It is based heavily on similar workflows in https://github.com/guardian/commercial/pull/797 and https://github.com/guardian/consent-management-platform/pull/849 (but with some slight differences):

1. adding the `🐥 Canaries` label to a PR will trigger the release of canary versions of all packages with waiting changesets[^1]
2. while the label is present, pushing changes to the PR head branch will also release a new set of canaries (so you don't have to remove and add it continually)
3. removing the label will stop new releases (until you add it back)

## Demo

You can see it running on this PR from https://github.com/guardian/csnx/pull/1178#event-11736477830 onwards.

[^1]: This is maybe not expected, but also it's not obvious how you'd get round this. How do you know which package with unreleased changes the user wants to test? It feels like overkill but I think it kind of makes sense.

### DIFF
--- a/.changeset/nice-walls-eat.md
+++ b/.changeset/nice-walls-eat.md
@@ -1,5 +1,0 @@
----
-'@guardian/libs': patch
----
-
-noop to test canary release workflow
--- a/.changeset/nice-walls-eat.md
+++ b/.changeset/nice-walls-eat.md
@@ -1,0 +1,5 @@
+---
+'@guardian/libs': patch
+---
+
+noop to test canary release workflow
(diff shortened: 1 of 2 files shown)


joecowton1 (approved):
nice work!

sndrs:
currently this adds all releases to git, its possible this may change https://github.com/changesets/action/issues/141 (which would be great)

PR #1184, opened 2024-02-08
Rationalise workflows a little

sndrs:
## What are you changing?

- moves the changesets and node/pnpm setup stuff to a new action
- makes sure storybook and packages are only published from `main` if CI succeeds

## Why?

- make it easier to read actions, and keep things in sync
- we shouldn't be releasing anything if `main` fails


oliverabrahams (approved):
Looks good

PR #1189, opened 2024-02-08
Move chromatic into a reusable workflow

sndrs:
## What are you changing?

- pulls chromatic into a reusable workflow
- uses it:
  - following a successful run of validate in CI
    - after a push to `main`
    - on PRs with the `run_chromatic` label
    - on PR from dependabot or changesets release bot
  - when a `run_chromatic` label is added to a PR  

## Why?

this should maintain the same level of chromatic coverage, but detaches CI from the labelling trigger in PRs


oliverabrahams (approved):
Looks good!

PR #1199, opened 2024-02-14
Display a file limit warning where applicable

abeddow91:
## What are you changing?
Adds a file limit warning underneath the supporting text on file input if a max file size has been provided.

## Why?
If there is a file size limit, this should be shown to the user before they try and upload a file. 

Screenshot 2024-02-14 at 10 53 51

### DIFF
--- a/libs/@guardian/source-react-components-development-kitchen/src/file-input/FileInput.stories.tsx
+++ b/libs/@guardian/source-react-components-development-kitchen/src/file-input/FileInput.stories.tsx
@@ -63,12 +63,7 @@ export const FileTypeValidation: StoryFn<typeof FileInput> = () => {
 
 export const FileSizeValidation: StoryFn<typeof FileInput> = () => {
 	return (
-		<FileInput
-			id="file-input"
-			label="Upload a file"
-			supporting="Only supports very very small images"
-			maxFileSize={1}
-		/>
+		<FileInput id="file-input" label="Upload a file" maxFileSize={5_000_000} />
 	);
 };
 
--- a/libs/@guardian/source-react-components-development-kitchen/src/file-input/FileInput.tsx
+++ b/libs/@guardian/source-react-components-development-kitchen/src/file-input/FileInput.tsx
@@ -9,6 +9,7 @@ import {
 	fileName as fileNameStyle,
 	fontSizes,
 	uploadSizes,
+	warningText,
 } from './styles';
 import type { Theme } from './theme';
 import type { FileInputProps } from './types';
@@ -87,6 +88,13 @@ export const FileInput: FC<FileInputProps> = ({
 				hideLabel={hideLabel}
 				cssOverrides={fontSizes[size]}
 			>
+				{maxFileSize && (
+					<p css={(theme: Theme) => warningText(theme.fileInput)}>
+						Please note, the maximum file size is{' '}
+						{getReadableFileSize(maxFileSize)}.
+					</p>
+				)}
+
 				{!!errorText && <InlineError>{errorText}</InlineError>}
 				<div
 					css={[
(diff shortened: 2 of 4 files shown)


libs/@guardian/source-react-components-development-kitchen/src/file-input/FileInput.tsx
```diff
@@ -12,6 +12,7 @@ import {
 } from './styles';
 import type { Theme } from './theme';
 import type { FileInputProps } from './types';
+import { SupportingText } from 'libs/@guardian/source-react-components/src/label/SupportingText';
```
mxdvl, line 15: I think this import is incorrect:

```suggestion
import { SupportingText } from '@guardian/source-react-components/src/label/SupportingText';
```

.changeset/twenty-panthers-run.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/source-react-components-development-kitchen': minor
```
mxdvl, line 2: I think this should be `major` as it can have a visual impact for consumers of the package.

libs/@guardian/source-react-components-development-kitchen/src/file-input/FileInput.stories.tsx
```diff
@@ -63,12 +63,7 @@ export const FileTypeValidation: StoryFn<typeof FileInput> = () => {
 
 export const FileSizeValidation: StoryFn<typeof FileInput> = () => {
 	return (
-		<FileInput
-			id="file-input"
-			label="Upload a file"
-			supporting="Only supports very very small images"
-			maxFileSize={1}
-		/>
+		<FileInput id="file-input" label="Upload a file" maxFileSize={5_000_000} />
```
mxdvl, line 66: Why override the existing story and not create a distinct one?

libs/@guardian/source-react-components-development-kitchen/src/file-input/FileInput.tsx
```diff
@@ -87,6 +88,13 @@ export const FileInput: FC<FileInputProps> = ({
 				hideLabel={hideLabel}
 				cssOverrides={fontSizes[size]}
 			>
+				{maxFileSize && (
+					<SupportingText>
+						Please note, the maximum file size is{' '}
+						{getReadableFileSize(maxFileSize)}.
+					</SupportingText>
+				)}
```
mxdvl, line 96: Do we still want to show this if there’s an error? It might feel quite redundant…

libs/@guardian/source-react-components-development-kitchen/src/file-input/FileInput.stories.tsx
abeddow91: The two stories wouldn't be distinct, they'd be the same here as both variants occur from a max file size being provided so an additional test would be a duplicate. I've removed the supporting text as it is redundant with the new warning message.

libs/@guardian/source-react-components-development-kitchen/src/file-input/FileInput.tsx
abeddow91: Hmm i'm not sure. There is one error which relates to the max file size but others don't relate and are separate to this message.

libs/@guardian/source-react-components-development-kitchen/src/file-input/FileInput.tsx
abeddow91: Yeah i thought the same but interestingly removing `libs/` results in this error https://github.com/guardian/csnx/actions/runs/7904047399/job/21573404262?pr=1199#step:5:370

mxdvl (approved):

PR #1217, opened 2024-02-19
Update `@guardian/browserslist-config` with latest data

sndrs:
## What are you changing?

- updating `@guardian/browserslist-config` with the latest data		

## Why?

- so we can be up-to-date


arelra (approved):
Thanks! Safari 10 and 11 still hanging in there 🥲

oliverabrahams (approved):

PR #1230, opened 2024-02-26
Add description to the Theme prop in storybook 

oliverabrahams:
## What are you changing?

- Add description of the colours you can change in storybook. 

## Why?

- Makes the intention of the prop clearer

Screenshot 2024-02-26 at 11 19 56
Screenshot 2024-02-26 at 11 20 15


libs/@guardian/source-react-components/src/accordion/Accordion.tsx
```diff
@@ -21,6 +21,19 @@ export interface AccordionProps extends Props {
...
+	 * Partial or complete theme to override the component's colour palette.
+	 * Sanctioned colours to change have been set out by the design system team.
+	 *
+	 *  The theme colours available to change are:
+	 *
+	 *  `textLabel`<br>
+	 *  `textBody`<br>
+	 *  `textCta`<br>
+	 *  `border`<br>
+	 *  `iconFill`
```
sndrs, line 34: it feels a bit brittle to have to keep these in sync with the theme?

it's not as elegant, but we set a default for the `theme` prop in the component, storybook can provide that in the docs:

image

how do you feel about this instead? (i'm not sure, but it's auto-generated which i do like)

jamesmockett: We'd still need to merge whatever is passed in via the `theme` prop with the default theme to handle partial themes so in a lot of cases we'd be merging the default theme with the default theme which feels like it could be confusing? (If we didn't allow partial themes we could assume the `theme` prop is either the default theme or another complete theme and avoid merging.)

I do like that it's auto-generated though. A similar approach is to set the `theme` prop in the story's args to the default theme, but then the `theme` prop is always set so non of the stories are testing the default scenario where it's undefined.

oliverabrahams: We wouldn't be able to do that because the theme prop has priority over the theme provider prop. So it would override any theme provided themes.

oliverabrahams: Another option is to do something like this 
Screenshot 2024-02-26 at 14 45 08
Screenshot 2024-02-26 at 14 45 29

oliverabrahams: The merged theme function is 
Screenshot 2024-02-26 at 15 13 15


So, If you have 

```
<ThemeProvider theme={background: 'blue'}
<Button/>
</ThemeProvider>
```

Then currently this would get through as the `themeOverrides` AKA `theme` prop are undefined. If they became the default theme by default. Then this background colour would be overwritten.

sndrs: yeah this seems too complex to pollute the intention of PR with. it would be great if this could somehow be automated but this is much better than right now, so let's merge this.

sndrs (approved):

PR #1232, opened 2024-02-26
testing turbosnap

oliverabrahams:
## What are you changing?

- Testing turbosnap. 

## Why?

- There is a new buildDir prop but it doesn't seem to fix the issues we have been having


sndrs (approved):
🤞 great digging

PR #1233, opened 2024-02-26
Remove base colour scale from tokens

jamesmockett:
## What are you changing?

- Removes base colour scale tokens. ie. `colour.red.1` etc.

## Why?

- These are not exported directly by Source Foundations and do not exist in Figma where the `palette` colours form the base of the colour system. (`brand.500` rather than `colour.blue.4`.) Having this additional layer of abstraction is potentially confusing and will cease to exist when the colour tokens are exported directly from Figma.

As these now form the base colours would it make sense to rename this token group from `palette` to `colour`? It can still be exported from Source Foundations as `palette` to avoid breaking existing references.


oliverabrahams (approved):
This is great 👍

PR #1240, opened 2024-02-28
Identity Auth | Optimise `isSignedInWithAuthState` to run only once at a time

coldlink:
## What are you changing?

-

## Why?

-


raphaelkabo (approved):
This is a very good check, well done. ✅

PR #1248, opened 2024-02-29
Map underline thickness tokens to font size

jamesmockett:
## What are you changing?

- Updates `text-decoration-thickness` tokens to map the desired value to specific font sizes as they don't change between font families. Currently this is configured for each font family and font size combination.

## Why

- This is an enabler to simplify implementation of the new web typography presets, but also makes the relationship between these values in the current typography API clearer.


sndrs:
does this add some opacity/inconsistency in what numbers mean in a name? 

a preset like `articleBoldItalic17` tells you the size is `17px`, but `textDecorationThickness[17]` give you a value of `2px`.

what do you think about something like `textDecorationThicknessForFontSize[]`? it's a bit cumbersome, but i think it keeps things clear and consistent

joecowton1 (approved):

PR #1273, opened 2024-03-08
Try out `wireit` on the storybook workflows

sndrs:
[Wireit](https://github.com/google/wireit) a lighter-weight alternative to Nx that we think could be better for this repo. This is test to see how it performs on non-critical stuff. If it does well, we'll expand it. If not, we'll revert to Nx.

Co-authored with @oliverabrahams.


jamesmockett (approved):
This is great. It's only a small corner of the repo, but Wireit feels a lot more familiar and easier to understand than Nx 👍

PR #1281, opened 2024-03-12
Include `--source-text-decoration-thickness` in preset CSS

jamesmockett:
## What are you changing?

Outputs `--source-text-decoration-thickness` custom property for each preset based on the font size.

## Why?

`--source-text-decoration-thickness` is used by the `<Link>` component to apply an appropriate text decoration style for links in the hover state.


oliverabrahams (approved):
Looks good

PR #1283, opened 2024-03-14
Add object literal representation of typography presets

jamesmockett:
## What are you changing?

Adds object literal representation of the web typography presets.

```ts
export const article15Object = {
  fontFamily: 'GuardianTextEgyptian, "Guardian Text Egyptian Web", Georgia, serif',
  fontSize: '0.9375rem',
  lineHeight: 1.4,
  fontWeight: 400,
  fontStyle: 'normal',
};
```


oliverabrahams (approved):
Looks Gd

mxdvl:
Would we want these objects to be marked `as const`, so their values are literals rather than primitives?

jamesmockett:
> Would we want these objects to be marked `as const`, so their values are literals rather than primitives?

I hadn't thought of that, but yes, I think that's a good idea 👍

PR #1289, opened 2024-03-18
Update the link component to use the new presets

oliverabrahams:
## What are you changing?

- Update the link component to use the new presets instead of having to use the css custom properties

## Why?

- When creating the new presets, it was a chance to remove the source specific css custom property


libs/@guardian/source-react-components/src/link/styles.ts
```diff
@@ -13,26 +14,28 @@ import type { IconSide, LinkPriority } from './@types/SharedLinkProps';
...
+export const link = (typographyPreset = textSans17) => {
+	return css`
+		position: relative;
+		cursor: pointer;
+		text-decoration: underline;
+		text-underline-position: under;
+		text-underline-offset: 5%;
+		${typographyPreset};
+		display: inline;
+		text-decoration-thickness: 0.1em;
```
jamesmockett, line 27: Could we nudge this up a couple of lines so it's grouped with the other text decoration properties?

PR #1291, opened 2024-03-18
Remove @guardian/source

joecowton1:
## What are you changing?

- removing the @guardian/source package as its in current state its blocking an update to NX, if we wish to continue trying to get that to work we can revert this PR


sndrs:
i think this is a good idea tbh. a single package would be great but there's no momentum behind it for now, so i say let's remove it until it becomes useful again (if it does)

.github/CODEOWNERS
```diff
@@ -1,26 +1,30 @@
 # If you would like to be notified about changes in particular part of this repo,
+
```
sndrs, line 2: is this meant to be part of this i wonder?

sndrs (approved):

PR #1304, opened 2024-03-25
Allow conversion of rem font size to pixels in typography presets

jamesmockett:
## What are you changing?

Adds `presetToPx` helper to convert default rem font size in a given typography preset to a pixel value. Line height is untouched as this is already specified as a unitless relative value.

## Why?

We want to encourage use of rem font sizes so rather than duplicating all of the presets with different units a conversion function provides a pixel equivalent where strictly necessary. 

```js
presetToPx(headlineBold24))
```

👇 

```css
font-family: "GH Guardian Headline", "Guardian Egyptian Web", Georgia, serif;
font-size: 24px;
line-height: 1.15;
font-weight: 700;
font-style: normal;
--source-text-decoration-thickness: 3px;
```

## Notes

This adds a `TypographyPreset` type to define the shape of a preset, but this hasn't been applied to the presets themselves as it hides the underlying values of the properties with the expected type:

Screenshot 2024-03-27 at 16 44 16

With untyped presets the values are exposed:

Screenshot 2024-03-27 at 16 43 52

### DIFF
--- a/libs/@guardian/source-foundations/src/utils/typography.ts
+++ b/libs/@guardian/source-foundations/src/utils/typography.ts
@@ -1,10 +1,15 @@
-const REGEX_FONT_SIZE = /font-size:\s(\d+\.\d+)rem/;
+/*
+ * Convert font size in typography preset from default rem value to pixels
+ */
 
 export const presetToPx = (preset: string) => {
+	const REGEX_FONT_SIZE = /font-size:\s(\d+\.\d+)rem/;
+
 	const matches = preset.match(REGEX_FONT_SIZE);
 	if (matches?.[1]) {
 		const pxVal = parseFloat(matches[1]) * 16;
 		return preset.replace(REGEX_FONT_SIZE, `font-size: ${pxVal}px`);
 	}
+
 	return preset;
 };
--- a/libs/@guardian/source-foundations/src/utils/typography.ts
+++ b/libs/@guardian/source-foundations/src/utils/typography.ts
@@ -1,0 +1,10 @@
+const REGEX_FONT_SIZE = /font-size:\s(\d+\.\d+)rem/;
+
+export const typographyPresetPx = (preset: string) => {
+	const matches = preset.match(REGEX_FONT_SIZE);
+	if (matches?.[1]) {
+		const pxVal = parseFloat(matches[1]) * 16;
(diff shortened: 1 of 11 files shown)


libs/@guardian/source-foundations/src/utils/typography.ts
```diff
@@ -0,0 +1,16 @@
...
+export const presetToPx = (preset: TypographyPreset) => {
+	const REGEX_FONT_SIZE = /font-size:\s(\d+\.\d+)rem/;
+
+	const matches = preset.match(REGEX_FONT_SIZE);
+	if (matches?.[1]) {
+		const pxVal = parseFloat(matches[1]) * 16;
+		return preset.replace(REGEX_FONT_SIZE, `font-size: ${pxVal}px`);
+	}
+
+	return preset;
```
mxdvl, line 15: You can also use a [replacer function as part of `String.prototype.replace`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_function_as_the_replacement), which could make this simpler:

```ts
return preset.replace(REGEX_FONT_SIZE, (_, match) => `font-size: ${parseFloat(match) * 16}px`)
```

libs/@guardian/source-foundations/src/utils/typography.ts
```diff
@@ -0,0 +1,16 @@
+import type { TypographyPreset } from '../typography/types';
+
+/*
+ * Convert `font-size` value in typography preset from rem to px
+ */
+export const presetToPx = (preset: TypographyPreset) => {
+	const REGEX_FONT_SIZE = /font-size:\s(\d+\.\d+)rem/;
```
mxdvl, line 7: I think you might have an issue if the value does not contain a decimal part…

```suggestion
	const REGEX_FONT_SIZE = /font-size:\s(\d+\.?\d*)rem/;
```

jamesmockett:
Thanks the comments @mxdvl and @SiAdcock 🙏  We've realised that people don't seem to be using the existing API's `unit: 'px'` option so given that we want to encourage the use of relative font sizes we've decided to leave this functionality out for now. We're not planning to deprecate the typography API immediately so if it turns out that there is a need for this we can resurrect the PR.

PR #1305, opened 2024-03-25
Add tertiary background to `ButtonTheme`

mxdvl:
## What are you changing?

- Add a tertiary background to `ButtonTheme`

## Why?

- This matches the [Source design guide](https://theguardian.design/2a1e5182b/p/435225-button)
- It would enable us to [drop CSS overrides for pagination buttons](https://github.com/guardian/dotcom-rendering/pull/11012/files#diff-ac87731b6529e69cd791781f11af90f9e8333a52077d154f61daf8cbf5f97657R111-R114)

### DIFF
--- a/libs/@guardian/source-react-components/src/button/theme.ts
+++ b/libs/@guardian/source-react-components/src/button/theme.ts
@@ -84,7 +84,7 @@ export const themeButton: ThemeButton = {
 	backgroundSecondary: palette.brand[800],
 	backgroundSecondaryHover: '#ACC9F7', // One-off colour variant generated from palette.brand[800]
 	textTertiary: palette.brand[400],
-	backgroundTertiary:  'transparent',
+	backgroundTertiary: 'transparent',
 	backgroundTertiaryHover: '#E5E5E5', // One-off colour variant
 	borderTertiary: palette.brand[400],
 	textSubdued: palette.brand[400],
--- a/libs/@guardian/source-react-components/src/button/theme.ts
+++ b/libs/@guardian/source-react-components/src/button/theme.ts
@@ -86,7 +86,7 @@ export const themeButton: ThemeButton = {
 	backgroundSecondary: palette.brand[800],
 	backgroundSecondaryHover: '#ACC9F7', // One-off colour variant generated from palette.brand[800]
 	textTertiary: palette.brand[400],
-	backgroundTertiary: palette.neutral[100],
+	backgroundTertiary:  'transparent',
 	backgroundTertiaryHover: '#E5E5E5', // One-off colour variant
 	borderTertiary: palette.brand[400],
 	textSubdued: palette.brand[400],
--- a/libs/@guardian/source-react-components/src/button/theme.ts
+++ b/libs/@guardian/source-react-components/src/button/theme.ts
@@ -9,7 +9,6 @@ export type ButtonTheme = {
 	backgroundSecondary?: string;
 	backgroundSecondaryHover?: string;
 	textTertiary?: string;
-	backgroundTertiary?: string;
 	backgroundTertiaryHover?: string;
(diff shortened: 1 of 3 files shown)


oliverabrahams (commented):
This is great! It would be really good to enable you to [drop CSS overrides for pagination buttons](https://github.com/guardian/dotcom-rendering/pull/11012/files#diff-ac87731b6529e69cd791781f11af90f9e8333a52077d154f61daf8cbf5f97657R111-R114)

libs/@guardian/source-react-components/src/button/theme.ts
```diff
@@ -36,7 +38,8 @@ export const buttonThemeDefault: { button: ButtonTheme } = {
 		textSecondary: palette.brand[400],
 		backgroundSecondary: palette.brand[800],
 		backgroundSecondaryHover: '#ACC9F7', // One-off colour variant generated from palette.brand[800]
-		textTertiary: palette.brand[400],
+		textTertiary: palette.neutral[100],
+		backgroundTertiary: palette.neutral[100],
```
oliverabrahams, line 42: Do we need to be updating this deprecated object.

libs/@guardian/source-react-components/src/button/theme.ts
```diff
@@ -9,6 +9,7 @@ export type ButtonTheme = {
 	backgroundSecondary?: string;
 	backgroundSecondaryHover?: string;
 	textTertiary?: string;
+	backgroundTertiary?: string;
```
oliverabrahams, line 12: I don't think we want to be updating this deprecated type.
```suggestion
```

libs/@guardian/source-react-components/src/button/theme.ts
```diff
@@ -83,6 +86,7 @@ export const themeButton: ThemeButton = {
 	backgroundSecondary: palette.brand[800],
 	backgroundSecondaryHover: '#ACC9F7', // One-off colour variant generated from palette.brand[800]
 	textTertiary: palette.brand[400],
+	backgroundTertiary: palette.neutral[100],
```
oliverabrahams, line 89: can we make this transparent to keep backwards compatibility ?
```suggestion
	backgroundTertiary:  'transparent',
```

libs/@guardian/source-react-components/src/button/theme.ts
oliverabrahams, line 42:
```suggestion
		textTertiary: palette.brand[400],
```

mxdvl: Not sure… it felt safer but will follow your lead!

libs/@guardian/source-react-components/src/button/theme.ts
mxdvl: The white is in Source!

oliverabrahams: Currently if someone is using it on a non white background this is will change it to white as this is the default. Like here in storybook. We could change it in the reader revenue ones to transparent or leave all of them as transparent.
Screenshot 2024-03-26 at 13 21 36

oliverabrahams: As in source it is always the same colour as the background I am inclined to leave this as transparent. Where in source are you referring to?

jamesmockett: It has a white background in the zeroheight docs, but this may be because the default theme is rendered against a grey background to differentiate it from the rest of the page:
https://theguardian.design/2a1e5182b/p/435225-button

In Figma the background colour is transparent.

oliverabrahams (approved):
Thanks for this @mxdvl. It will be great to see the themes being used to replace CSS overrides!

PR #1306, opened 2024-03-26
Type preset objects `as const`

mxdvl:
## What are you changing?

- Mark the type objects as const

## Why?

- Suggested in https://github.com/guardian/csnx/pull/1283#issuecomment-2020031105


jamesmockett (approved):
Thanks! 🙏

PR #1330, opened 2024-04-03
Add deprecated export file and relocate deprecated code

jamesmockett:
## What are you changing?

- Relocates deprecated code to `_deprecated` folder and adds exports to `deprecated-exports.ts`.
- Removes deprecated `resets.input` styles completely as these are now unused.

## Why?

- Having all deprecated code and exports in a single place makes it simpler to remove in the future. `deprecated-exports.ts` already exists in other packages so this is replicating the same pattern.


libs/@guardian/source-foundations/src/_deprecated/typography/types.ts
```diff
@@ -0,0 +1,4 @@
+/** @deprecated Type will be removed in a future major release */
```
sndrs, line 1: is this message correct? the export is `TypographySizes`?

sndrs, line 1: is this message correct? should it be `TypographySizes`?

```suggestion
/** @deprecated TypographySizes will be removed in a future major release */
```

sndrs (approved):

PR #1333, opened 2024-04-04
Refine `wireit` storybooks setup

sndrs:
## What are you changing?

- uses `pnpm` to run all storybooks for the composed one

## Why?

- less config to get right, and improves the terminal output:

image


oliverabrahams (approved):
Nice!

PR #1334, opened 2024-04-04
Move Nx configs from `project.json` to `package.json`

sndrs:
## What are you changing?

Moving all project-level Nx config from `project.json` files to the project's `package.json`.

## Why?

`project.json` is too Nx-specific for the unfamiliar, and doesn’t offer us anything, since we’re very much in a Node world.

It also sets up an easier transition to `wireit` (`wireit` also puts its config in `package.json`.


oliverabrahams (approved):
Looks good to me

PR #1369, opened 2024-04-18
Improve `Inline` layout spacing

jamesmockett:
## What are you changing?

Removes additional spacing from bottom of child elements so that they are flush with the container

## Why?

The `Inline` layout component is used to lay out child elements horizontally with consistent spacing added between adjacent children. Children will wrap on to another row if there isn't enough horizontal spacing. If wrapping occurs the same spacing is applied between rows. Children should otherwise sit flush with the edges of the container. Currently spacing is added to the bottom of every child which is particularly noticeable when larger spacing values are used.

We could potentially use the [`gap` property](https://developer.mozilla.org/en-US/docs/Web/CSS/gap) in place of negative margins, although this [would exclude older browsers](https://caniuse.com/?search=gap). Unfortunately support for `gap` on flex layouts cannot be reliably feature detected as older browsers may support `gap` for grid layout, but not flex layouts.

### Before

Screenshot 2024-04-18 at 12 22 15

Screenshot 2024-04-18 at 12 22 32

### After

Screenshot 2024-04-18 at 16 29 19

Screenshot 2024-04-18 at 16 29 31

Screenshot 2024-04-18 at 16 29 41

### DIFF
--- a/libs/@guardian/source-react-components/src/@types/Space.ts
+++ b/libs/@guardian/source-react-components/src/@types/Space.ts
@@ -1,0 +1,3 @@
+import type { space } from '@guardian/source-foundations';
+
+export type Space = keyof typeof space;
--- a/.changeset/early-rocks-run.md
+++ b/.changeset/early-rocks-run.md
@@ -1,0 +1,5 @@
+---
+'@guardian/source-react-components': major
+---
+
+Updates `Inline` layout component to remove additional spacing from bottom of child elements so they sit flush with the edges of the container
--- a/libs/@guardian/source-react-components/src/inline/@types/InlineSpace.ts
+++ b/libs/@guardian/source-react-components/src/inline/@types/InlineSpace.ts
@@ -1,1 +1,0 @@
-export type InlineSpace = 1 | 2 | 3 | 4 | 5 | 6 | 9 | 12 | 24;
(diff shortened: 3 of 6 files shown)


libs/@guardian/source-react-components/src/@types/Space.ts
```diff
@@ -0,0 +1,3 @@
+import type { space } from '@guardian/source-foundations';
+
+export type Space = keyof typeof space;
```
sndrs, line 3: could we use the export from `@guardian/source-foundations` directly? i'm not sure what this gives us?

sndrs (approved):
great apart from the query about the `space` type

libs/@guardian/source-react-components/src/@types/Space.ts
jamesmockett: This is replacing the component's own hardcoded list of spacing values:

```ts
export type InlineSpace = 1 | 2 | 3 | 4 | 5 | 6 | 9 | 12 | 24;
```

A couple of other components have a similar spacing type, but they're all missing values that have been subsequently added.

`@guardian/source-foundations` doesn't export a spacing type so you'd have to use `keyof typeof space` to get these. Foundations would be a better place for this though.

Perhaps I should add this to foundations and update all of the components to use it in a separate PR?

sndrs: sorry james, i was too hasty reviewing. i think this is the right approach, sorry for the noise

PR #1374, opened 2024-04-24
Requires curly braces in all circumstances in the eslint config

sndrs:
## What are you changing?

- make curly braces compulsory

## Why?

- reduce noise in diffs
- remove ambiguity about when you should use curly braces
- (possibly) remove ambiguity about when a blocks start and end (especially for people unfamiliar with the language)


arelra (approved):
🙏

libs/@guardian/ab-core/src/fixture.ts
```diff
@@ -42,7 +42,9 @@ export const genAbTest = (genAbConfig: GenAbConfig): ABTest => {
 		audience,
 		author: 'n/a',
 		canRun: (): boolean => {
-			if (canRun !== undefined) return !!canRun;
+			if (canRun !== undefined) {
+				return !!canRun;
+			}
```
mxdvl, line 47: This is quite the mouthful… will address in a follow-up PR

libs/@guardian/libs/src/consent-management-platform/index.ts
```diff
@@ -20,7 +20,9 @@ import { initVendorDataManager } from './vendorDataManager';
 // Store some bits in the global scope for reuse, in case there's more
 // than one instance of the CMP on the page in different scopes.
 if (!isServerSide) {
-	if (typeof window.guCmpHotFix === 'undefined') window.guCmpHotFix = {};
+	if (typeof window.guCmpHotFix === 'undefined') {
+		window.guCmpHotFix = {};
+	}
```
mxdvl, line 25: This can be folded in the parent if… will address in a follow-up!

libs/@guardian/libs/src/datetime/timeAgo.ts
```diff
@@ -65,33 +65,55 @@ export const timeAgo = (
...
-			if (length == 1) return '1 minute ago';
+			if (length > 55) {
+				return verbose ? '1 hour ago' : '1h ago';
+			}
+			if (!verbose) {
+				return `${length}m ago`;
+			}
+			if (length == 1) {
+				return '1 minute ago';
+			}
```
mxdvl, line 94: I think these would look nicer with an `else if`, now… might address in a follow-up

mxdvl (approved):
Great, this also ensures that [there’s no conflict between ESLint and Prettier](https://github.com/prettier/eslint-config-prettier#curly)!

libs/@guardian/libs/src/datetime/timeAgo.ts
mxdvl: #1376

PR #1377, opened 2024-04-24
Refactor `ab-core` for clearer intent

mxdvl:
## What are you changing?

- use nullish coalescing to convert a `boolean | undefined` into a `boolean`

## Why?

- nullish coalescing is shorter and more expressive about the intent
- a [follow-up on a recent formatting change](https://github.com/guardian/csnx/pull/1374#discussion_r1577925315)


sndrs (approved):
nice one - can you add a changeset?

PR #1380, opened 2024-04-25
Update Design Tokens package to export multiple objects

oliverabrahams:
## What are you changing?

- Updates _Design Tokens_ package to export separate objects for the different types of token. e.g. palette, typography, spacing etc.

## Why?

- Currently all tokens are exported as a single object which can't be tree shaken, leading to some bloat in _Source Foundations_ where it's having to import more JS than required.


libs/@guardian/source-foundations/src/colour/palette.ts
```diff
@@ -1,24 +1,8 @@
...
-	sport: tokens.palette.sport,
-	culture: tokens.palette.culture,
-	lifestyle: tokens.palette.lifestyle,
-	labs: tokens.palette.labs,
-	specialReport: tokens.palette.specialReport,
-	specialReportAlt: tokens.palette.specialReportAlt,
-	focus: tokens.palette.focus,
-} as const;
+
+export { palette } from '@guardian/design-tokens';
```
jamesmockett, line 8: ❤️

oliverabrahams: I remember why we did this. It was for the ordering in chromatic

jamesmockett (approved):

libs/@guardian/source-foundations/src/colour/palette.ts
jamesmockett: Ah, yeah, that was probably it. I don't think the ordering in Chromatic matters so long as the colours are correct. This is a lot more concise than constructing a new palette object.

PR #1385, opened 2024-04-26
Simplify the build pipeline

sndrs:
## What are you changing?

updates the build/packages setup:

- all packages are ES modules
- builds them with `rollup` directly
  - removes `@csnx/npm-package`
- uses `rollup-plugin-esbuild` and `rollup-plugin-dts` in place of `rollup-plugin-ts`
- builds to local `dist` dirs and removes the `paths` config from `tsconfig`
- removes unnecessary `build` and `validate-dist` tasks from packages that don't actually build anything (were just being copied to Nx's root `dist` dir)

## Why?

All packages that depend on sibling packages in the repo now consume their real `dist` files, not `.ts` source files.

it also surfaces the rollup config and package.json for each package more clearly.

our previous typescript plugin created some weird type issues (we think it was the plugin). `rollup-plugin-dts` is more commonly used hopefully more robust!

## Why is it so big?

we'd accumulated some tech debt that made each of these steps difficult to do alone - this was pairing/mobbing effort the whole team was involved in, so in effect it was reviewed in real time.

Co-authored-by: @oliverabrahams 
Co-authored-by: @jamesmockett


libs/@guardian/ab-core/tsconfig.json
```diff
@@ -6,5 +6,8 @@
 		{
 			"path": "./tsconfig.spec.json"
 		}
-	]
+	],
+	"compilerOptions": {
+		"incremental": false
```
sndrs, line 11: this comes from `@guardian/tsconfig`, which feels wrong to me

libs/@guardian/ab-react/tsconfig.json
```diff
@@ -1,5 +1,7 @@
 {
 	"extends": "../../../tsconfig.base.json",
-	"files": [],
-	"include": ["."]
+	"include": ["."],
+	"compilerOptions": {
+		"tsBuildInfoFile": "tsconfig.tsbuildinfo"
+	}
```
sndrs, line 6:
```suggestion
	"include": ["."]
```
lets remove incremental from our tsconfig

libs/@guardian/libs/tsconfig.json
```diff
@@ -6,5 +6,13 @@
...
 			"path": "./tsconfig.spec.json"
 		}
-	]
+	],
+	"compilerOptions": {
+		"incremental": false
+	},
+	"ts-node": {
+		"moduleTypes": {
+			"jest.config.ts": "cjs"
```
sndrs, line 15: ?

libs/@guardian/source-foundations/src/typography/stories/typography.mdx
```diff
@@ -97,4 +97,4 @@ _Note:_ this font is not available on [theguardian.com](https://theguardian.com)
 ### Specimen
 
 <Canvas of={TypographyStories.Presets} />
-import {textEgyptian17} from 'libs/@guardian/source-foundations/vendor/typography/css';
```
jamesmockett, line 100: This looks like an instance of VS Code automatically adding an import, but can be safely removed as it's not meant to be there.

sndrs: the whole import?

jamesmockett: Yes, the whole import. The `<Canvas>` element _should_ be the last thing in that file.

jamesmockett: Good timing with my reply as you've literally just done that 😅

oliverabrahams (approved):

PR #1386, opened 2024-04-28
`Nx` → `wireit`

sndrs:
## What are you changing?

swap out `Nx` for `wireit`

## Why?

it's much more lightweight, and is only concerned orchestrating tasks, rather than providing deps, generators, plugins etc.

we think it will make it easier to manage the repo and present a far smaller learning curve to new users of the repo


libs/@guardian/source-react-components-development-kitchen/package.json
```diff
@@ -54,65 +54,73 @@
...
+			"files": [
+				"../../../configs/rollup.config.js",
+				"./rollup.config.js",
+				"src/**",
+				"package.json",
+				"tsconfig.json"
+			],
+			"output": [
+				"dist"
+			]
```
oliverabrahams, line 69: Does this need a dependancy on react-components and foundations build

libs/@guardian/source-react-components/package.json
```diff
@@ -66,58 +66,68 @@
...
-					"{projectRoot}/**",
-					"{workspaceRoot}/jest.*"
-				]
-			}
-		}
-	},
 	"wireit": {
+		"build": {
+			"command": "rollup -c",
+			"files": [
```
oliverabrahams, line 72: Does this need a dependency on foundations build

libs/@guardian/source-react-components-development-kitchen/package.json
```diff
@@ -54,65 +54,73 @@
...
+		},
 		"build-storybook": {
```
oliverabrahams, line 71: And here?

oliverabrahams (commented):
Looks good! added some comments about dependencies

libs/@guardian/source-react-components-development-kitchen/package.json
sndrs: yeah good shout

oliverabrahams (approved):

PR #1391, opened 2024-04-29
default consents for 11 is false

vlkerag:
## What does this change?
Default consent 11 to false. Note that this will set canTarget to false if only purpose 11 is de-selected which is not currently the case.
Copy over from https://github.com/guardian/consent-management-platform/pull/868

## Why?
It is currently 'undefined' if reader rejected all or explicitly rejected purpose 11


akinsola-guardian (approved):
LGTM.

PR #1392, opened 2024-04-29
Do not enable `incremental` in `@guardian/tsconfig`

sndrs:
## What are you changing?

removes `"incremental": true` from `@guardian/tsconfig`

## Why?

It should be a project's own choice to use it, if it's useful to that project.

It's not a valuable default, as [it requires other settings to work well](https://www.typescriptlang.org/tsconfig/#incremental).

### DIFF
--- a/.changeset/wet-llamas-camp.md
+++ b/.changeset/wet-llamas-camp.md
@@ -1,0 +1,7 @@
+---
+'@guardian/tsconfig': minor
+---
+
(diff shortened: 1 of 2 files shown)


mxdvl (approved):
Seems fair to consider this a repo-level consideration.

[Working without issues it did not](https://github.com/guardian/csnx/pull/225#issuecomment-1326638897)…

Is it me or with each new version of TS it feels like the combinations get ever more complex? I’ve got [this cheat sheet to go through](https://www.totaltypescript.com/tsconfig-cheat-sheet) which seems to highlight more recent changes.

PR #1394, opened 2024-04-30
Update Source Development Kitchen components to use typography presets

jamesmockett:
## What are you changing?

- Updates Source Development Kitchen components to latest version of Source Foundations and replaces uses of typography API with typography presets.

## Why?

- The typography API is deprecated and will be removed in a future release.

### DIFF
--- a/.changeset/old-zebras-divide.md
+++ b/.changeset/old-zebras-divide.md
@@ -1,0 +1,5 @@
+---
+'@guardian/source-react-components-development-kitchen': major
+---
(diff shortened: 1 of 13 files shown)


oliverabrahams:
Do you want to add a changeset to this one also?

jamesmockett:
> Do you want to add a changeset to this one also?

I do, yes 😄 Also need to update the `@guardian/source-react-components` now that has been updated.

PR #1404, opened 2024-05-01
Use `bundler` for module resolution in `tsconfig.json`

sndrs:
## What are you changing?

move the TS module resolution algorithm from `node` (i.e. `node10`) to `bundler`.

it differs from `nodenext` in that [it does not require file extensions on relative paths](https://www.typescriptlang.org/tsconfig/#moduleResolution)

## Why?

1. it enables our packages to be consumed in environments that make use of `package.json#exports` (currently they cannot)
2. the old way is very outdated, and this enables writing ES modules in TypeScript that import `.ts` in `src` but eventually load the imports as transpiled `.js` when run. 
4. using `package.json#exports` internally also means we can only consume dependencies via their public API, rather than (as currently) importing things which are really only the side-effects of the author's implementation


tsconfig.base.json
```diff
@@ -45,7 +45,8 @@
 			]
 		},
 		"rootDir": ".",
-		"sourceMap": true
+		"sourceMap": true,
+		"moduleResolution": "bundler"
```
mxdvl, line 49: Nice!

mxdvl (approved):

oliverabrahams (approved):

PR #1408, opened 2024-05-02
Remove deprecated input reset style

jamesmockett:
## What are you changing?

- Removes deprecated input reset style.

## Why?

- This style is already part of the styles for the Source input components that need it so the separate export is unused.
- We previously moved all deprecated code into the `_deprecated` folder, but were unable to do so with this style as it's not exported separately, but as part of a larger object which is not deprecated. By removing this unused code _all_ deprecated code is now in a single location.

### DIFF
--- a/.changeset/shy-buckets-obey.md
+++ b/.changeset/shy-buckets-obey.md
@@ -1,0 +1,5 @@
+---
+'@guardian/source-foundations': major
+---
+
+Removed deprecated input reset style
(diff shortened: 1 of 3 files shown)


sndrs (approved):
be gone

.changeset/shy-buckets-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/source-foundations': major
+---
+
+Removed deprecated input reset style
```
sndrs, line 5:
```suggestion
Removed deprecated `resets.input` reset style.
```

PR #1409, opened 2024-05-02
Fix CMP cypress tests

sndrs:
## What are you changing?

update the CMP e2e to expect the current version of the CMP

## Why?

the CMP changed, so the cypress test broke

longer term, we should not really be testing for the presence of text in a button in document on the web that is not affected by code in this repo, but for now this stops the tests failing and allows us to publish packages again.


akinsola-guardian (approved):
Thank you for doing this.

PR #1412, opened 2024-05-03
Remove return types for react/emotion components

sndrs:
## What are you changing?

removes all return types for react/emotion components.

## Why?

`JSX.Element` is maximally wide, and so possibly incorrect. TypeScript can infer the correct (possibly narrower - e.g. it doesn’t accept children) type.

See https://github.com/guardian/dotcom-rendering/pull/5622 for more info.


jamesmockett (approved):
🚀

PR #1433, opened 2024-05-08
Use common `Space` type across components

jamesmockett:
## What are you changing?

- Use the common `Space` type in components that accept spacing values as props.
- Fixes a bug where spacing styles are not applied if `space.0` is used due to `0` being falsy.
- Update `Stack` stories to better show how spacing is applied.
- Expands `Inline` and `Stack` stories to include examples of _all_ spacing values.

Screenshot 2024-05-08 at 15 00 27

## Why?

- It removes a small amount of duplication and avoids having to update spacing values in multiple places. [New spacing values were added last year](https://github.com/guardian/csnx/pull/897), but some components still lack support for these as their types were not updated at the same time.


sndrs (approved):
looks good, does it need a patch release?

jamesmockett:
> looks good, does it need a patch release?

Good point. I was thinking this doesn't need a release, but I suppose it _does_ expose some new spacing values in those components. I'll add a changeset…

libs/@guardian/source/package.json
```diff
@@ -40,6 +40,7 @@
 		"@babel/core": "7.24.0",
 		"@emotion/react": "11.11.1",
 		"@guardian/design-tokens": "workspace:*",
+		"@guardian/libs": "workspace:*",
```
jamesmockett, line 43: Should I use the workspace version of `libs`, or specify a specific version as in `source-react-components`?

sndrs: if we use the workspace version, it means needing to bump this package everytime `libs` changes, so i'd be inclined to not so that we can avoid that.

jamesmockett: I've updated to match `source-react-components`.

sndrs (approved):

PR #1435, opened 2024-05-09
Try `@guardian/source` package once more

sndrs:
## What are you changing?

create a single `@guardian/source` package (again).

> [!WARNING]
> This PR does not get rid of the old ones yet. They should still be what people use. This version could go away again pretty quickly.

## Why?

managing `peerDeps` between tightly bound packages is hard, especially when they live in he same repo.

ideally you want the code on the disk to be the code that runs i.e. if you change `source-foundations` you want to see that reflected in `source-react-components` while developing. 

but that means keeping the `peerDep` range of `source-react-components` in sync with the current version of `source-foundations`. 

This in turn means releasing a major version of `source-react-components` everytime there's a new version of `source-foundations`, even if it's a patch.

This is the _correct_ thing to do, but it's not _desirable_. 

We concluded that was a smell coming from the package set up. The Source packages are tightly coupled, and the reason they are separate now is because we didn't understand properly how to support the old Node module resolution algorithm with sub-packages. We also didn't know about [`peerDependenciesMeta`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#peerdependenciesmeta) last time we looked at it.

Using those, we can create a single Source package that maps:

- `@guardian/source-foundations` → `@guardian/source/foundations`
- `@guardian/source-react-components` → `@guardian/source/react-components`

We can also make React/emotion optional `peerDeps` for people not using the react components.

Most package managers will install peerdeps anyway, so the optional flag just ensures that any version a consumer may also install matches the one we need.

---
Co-authored with @oliverabrahams


jamesmockett (approved):
📦📦→📦 🚀

PR #1447, opened 2024-05-15
Update generated id with react useId 

oliverabrahams:
## What are you changing?

- swap generated id with react useId

## Why?

- Inside React, useId is generated from the “parent path” of the calling component.
- https://github.com/guardian/dotcom-rendering/pull/11420


SiAdcock (approved):
💎 💎 💎

PR #1469, opened 2024-05-22
force pnpm to use registry version of packages

oliverabrahams:
## What are you changing?

- Update lockfile and .npmrc to use workspace versions of packages. pnpm v8 does not do this by default. 

## Why?

-After moving to wireit and removing the paths from TSconfig. Pnpm was now linking the workspace versions of packages if they were the same. This meant that you had to build the workspace packages to get the dependant projects to work. 

Screenshot 2024-05-22 at 10 29 57

- Either we add a `pnpm -r build` to the `make install` cmd or do this?


sndrs (approved):
great

PR #1475, opened 2024-05-22
Remove source eslint plugins

sndrs:
## What are you changing?

Removing the source eslint plugins from the repo.

## Why?

They've not been updated since they automated the move from `v3` → `v4` and don't check anything else. 

They also don't apply the new single source package.


jamesmockett:
There are a couple of references in CODEOWNERS that could be removed too.

SiAdcock (approved):

PR #1476, opened 2024-05-22
Promote spinner to standalone component

jamesmockett:
## What are you changing?

- Promotes spinner from icon library to become a standalone component
- Updates default colours and simplifies markup
- Fixes animation in Safari
- Adds theming support
- Updates `Button` component to use spinner theme rather than overriding styles

|Before|After|
|---|---|
| Screenshot 2024-05-21 at 16 43 38 | Screenshot 2024-05-21 at 16 44 00 |

`#a5c0e8` → `palette.brand[800]`

## Why?

- The spinner has different requirements to the other icons and is not generated as part of the icon build script
- It's using a deprecated palette export and hardcoded non-Source colour
- The animated circle is marked up as a `path` element rather than a simpler `circle` 
- Circle radius is a fraction too larger leading to clipping at the edge when shown at larger sizes
- We want to avoid style overrides unless strictly necessary and already have a theming system in place to support this

## Additional information

The `Size` prop has been extended to support custom sizes in addition to the existing icon sizes. (These have been retained for compatibility with existing uses of `SvgSpinner`.)

```tsx
<>
  <Spinner size="small" />
  <Spinner size={40} />
</>
```

Buttons with `isLoading` enabled use custom spinner sizes. Extending the `size` prop to support custom sizes allows this to be done without the need to override styles as is currently necessary. Updating the button component has _not_ been done in this PR as it requires updating all of the icon styles which is beyond the scope of this PR. (Without these changes any custom size set with the `size` prop is overridden again by the button's default icon size overrides.)

## To do

- [x] Colour change approved by Design System team
- [x] Add more size options to support use in buttons without needing to override styles
- [x] Promote spinner to a standalone component rather than part of the icon library
- [x] Add documentation and guidance around accessibility


libs/@guardian/source/src/react-components/spinner/Spinner.stories.tsx
```diff
@@ -0,0 +1,65 @@
...
+	size: 40,
+};
+
+// *****************************************************************************
+
+export const CustomTheme: StoryFn<typeof Spinner> = Template.bind({});
+CustomTheme.args = {
+	theme: {
+		strokeBackground: 'transparent',
+		strokeForeground: palette.neutral[7],
```
sndrs, line 63: what do you think about these names for the theme props? they sound more natural to me?

```suggestion
		background: 'transparent',
		colour: palette.neutral[7],
```

jamesmockett: The awkward names are partly because there are also corresponding `fillBackground` and `fillForeground` props. Thinking about it, I don't see any reason to ever use those so I could remove them and simplify the theme down to `background` and `colour` which does seem more obvious 👍

PR #1489, opened 2024-05-24
Make `react` and `@emotion/react` optional `peerDeps` of source

sndrs:
## What are you changing?

make `react` and `@emotion/react` optional `peerDeps` of source

## Why?

you don't need them if you're not using react components and i've just noticed we haven't done it!


jamesmockett (approved):
📦

PR #1494, opened 2024-05-24
Sort `package.json` in new Source packages

jamesmockett:
## What are you changing?

- Sorted `package.json` files in new Source packages by running `sort-package-json`.

## Why?

- `sort-package-json` runs as a pre-commit hook. Merging from `main` into branches that existed before these new packages were created may cause these changes to be applied there depending on what files have been touched. It makes sense to apply these changes in a single place rather than an unrelated branch.

### DIFF
--- a/libs/@guardian/source-development-kitchen/react-components/package.json
+++ b/libs/@guardian/source-development-kitchen/react-components/package.json
@@ -1,6 +1,6 @@
 {
-	"//": "sub-package to match the `exports` field in the parent package.json",
 	"private": true,
 	"type": "module",
-	"main": "../dist/react-components.js"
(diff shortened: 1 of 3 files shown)


jamesmockett:
Although we probably want the comment as the first thing in the file. Should we exclude these from the pre-commit hook instead?

libs/@guardian/source-development-kitchen/react-components/package.json
```diff
@@ -1,6 +1,6 @@
 {
-	"//": "sub-package to match the `exports` field in the parent package.json",
```
sndrs, line 2: oh this is annoying...

PR #1497, opened 2024-05-28
Add `CONTAINER` to component event type

cemms1:
## What are you changing?

- Adds `"CONTAINER"` to `OphanComponentEvent` type

## Why?

- [This already exists in Ophan](https://github.com/guardian/ophan/blob/20d491fe98095cb708bb3e4ff90b33568ea97c72/event-model/src/main/thrift/componentevent.thrift#L209)
- To allow web to send component events about front containers via Ophan

### DIFF
--- a/.changeset/kind-pots-fly.md
+++ b/.changeset/kind-pots-fly.md
@@ -1,0 +1,5 @@
+---
+'@guardian/libs': patch
+---
+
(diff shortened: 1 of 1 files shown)


mxdvl:
Hopefully this package is soon superceded by `ophan-tracker-js` providing its own types…

PR #1501, opened 2024-05-30
Set `moduleResolution` to `bundler` in `@guardian/tsconfig`

sndrs:
## What are you changing?

Setting `moduleResolution` to `bundler` in `@guardian/tsconfig`.

## Why?

It enables support for `exports` in TypeScript while not requiring strict import names (using extensions) i.e. it mimics how bundlers approach imports from local and installed modules. 

See https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/#moduleresolution-bundler for more info.


.changeset/silly-rivers-allow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/tsconfig': minor
```
oliverabrahams, line 2: Is this not a major change as it will mean that consumers may have to change their code?

sndrs: discussed offline - it is breaking but `0.*.*` version changes can all be breaking, until you go to the first major (`1.0.0`).

however, we decided should just be v1 anyway, so will make it a major.

oliverabrahams (approved):
💯

PR #1507, opened 2024-06-03
Add AB test to use `propertyId` in CMP sourcepoint config

Jakeii:
## What are you changing?
Testing the suggested change from [sourcepoint documentation](https://docs.sourcepoint.com/hc/en-us/articles/8938401981843-Best-practices-propertyHref#h_01GB81HYASFG9GJ4Q3GR0VSQK2), to use `propertyId` rather than `propertyHref`

Most of the world uses a single id, but Australia has their own for some reason.

## Why?
When using `propertyHref` sourcepoint uses an extra network request to determine consent, currently there's 2-3 requests depending on if there are any changes on sourcepoint's side i.e. to the vendorlist. After this change it will be reduced to 1-2.

This should speed up getting consent by 50-33% as it appears to be mostly network bound, and therefore anything else that waits for consent, e.g. advertising.

### Before:
Screenshot 2024-05-23 at 13 30 02

### After:
Screenshot 2024-05-23 at 13 29 33


emma-imber (approved):
Looks good from a commercial perspective!

libs/@guardian/libs/src/consent-management-platform/types/window.d.ts
```diff
@@ -31,8 +31,8 @@ declare global {
 			config: {
 				baseEndpoint: EndPoint;
 				accountId: number;
-				propertyHref: Property;
-				propertyId?: string;
+				propertyHref?: Property;
+				propertyId?: number;
```
sndrs, line 35: is this a breaking change?

Jakeii: No change to public APIs

sndrs: 👍

akinsola-guardian (approved):
LGTM. Thank you.

mxdvl (approved):

sndrs (approved):

PR #1510, opened 2024-06-04
migrate from cypress to playwright

oliverabrahams:
## What are you changing?

- Replaced the cypress e2e tests with playwright

## Why?

- The cypress ones were quite slow and flakey


libs/@guardian/libs/playwright.config.js
```diff
@@ -0,0 +1,30 @@
...
+	},
+	projects: [
+		{
+			name: 'chromium',
+			use: { ...devices['Desktop Chrome'] },
+		},
+	],
+
+	webServer: {
+		command: 'cd ../../../ && pnpm --filter github-pages start',
```
sndrs, line 26: do you need to cd up here?

oliverabrahams: Good point! Removed it

libs/@guardian/libs/package.json
```diff
@@ -69,13 +68,23 @@
...
+				"!.eslintcache",
+				"!jest*",
+				"!(playwright-report)/**",
+				"!(test-results)/**",
+				"!*.md"
 			],
-			"output": [],
+			"output": []
+		},
+		"e2e:ui": {
```
jamesmockett, line 86: Does this need ` files` and `output` defined like the `e2e` task?

jamesmockett (approved):
Other than the question about the `e2e:ui` task this looks good. Nice work 🎉

libs/@guardian/libs/package.json
oliverabrahams: I dont think so because they are for caching. I don't think we want to cache this one as it is a debug mode for the tests - https://github.com/google/wireit/blob/main/README.md#caching

jamesmockett: Ah, yes, that makes sense! 😄

sndrs (approved):
so much nicer! great work

PR #1513, opened 2024-06-05
First pass at documenting layout components and patterns

jamesmockett:
## What are you changing?

- Adds doc page to Source storybook to document layout components and show examples

## Why?

- The layout components in Source are not that widely used. This is an attempt to begin documenting them and show how they can be composed to create common layout patterns with consistent spacing


oliverabrahams (approved):
This is great. maybe one for the client side infra

jamesmockett:
> This is great. maybe one for the client side infra

The docs and some of the layout components need a little refinement still, but definitely something worth discussing in a future Client Side Meetup.

PR #1564, opened 2024-06-19
getConsentFor handle unknown ids

Jakeii:
## What are you changing?
move `google-analytics` to a new part of `vendorDataManager` that will remove its keys even if the id is no longer in the vendorlist.

This is currently throwing an error as the `vendorDataManager` checks for consent before removing a key.

This also narrows the type of `getConsentFor` and the vendorlist to catch errors in the future.

## Why?
If we remove a vendor we still want to be removing it's data from cookies/localStorage.

### DIFF
--- a/libs/@guardian/libs/src/consent-management-platform/getConsentFor.ts
+++ b/libs/@guardian/libs/src/consent-management-platform/getConsentFor.ts
@@ -8,7 +8,7 @@ export const getConsentFor: GetConsentFor = (
 ): boolean => {
 	const sourcepointIds = VendorIDs[vendor];
 
-	if (typeof sourcepointIds === 'undefined' || Array.isArray(sourcepointIds)) {
+	if (typeof sourcepointIds === 'undefined' || sourcepointIds.length === 0) {
 		throw new Error(
 			`Vendor '${vendor}' not found, or with no Sourcepoint ID. ` +
 				'If it should be added, raise an issue at https://github.com/guardian/consent-management-platform/issues',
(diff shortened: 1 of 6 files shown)


akinsola-guardian (approved):
Thanks. Looks good.

libs/@guardian/libs/src/consent-management-platform/getConsentFor.ts
```diff
@@ -8,10 +8,7 @@ export const getConsentFor: GetConsentFor = (
 ): boolean => {
 	const sourcepointIds = VendorIDs[vendor];
```
mxdvl, line 9: If you wanted to stop checking whether this was defined you could do a nullish coalescing to an empty array:

```ts
const sourcepointIds = VendorIDs[vendor] ?? [];
```

libs/@guardian/libs/src/consent-management-platform/vendorDataManager.ts
```diff
@@ -3,31 +3,42 @@ import { storage } from '../storage/storage';
...
-import type { VendorWithData } from './vendorStorageIds';
-import { vendorStorageIds } from './vendorStorageIds';
+import type { VendorData, VendorWithData } from './vendorStorageIds';
+import {
+	deprecatedVendorStorageIds,
+	vendorStorageIds,
+} from './vendorStorageIds';
 
-const removeData = (consent: ConsentState) =>
+const removeData = (vendorData: VendorData): void => {
```
mxdvl, line 12: By using destructuring here and setting defaults you can safely loop over each array rather than having to check for their existence first:

```suggestion
const removeData = ({
  cookies = [],
  localStorage = [],
  sessionStorage = [],
}: VendorData): void => {
```

mxdvl (commented):
Some thoughts on what a default looks like for arrays — I think `[]` is clearer than `undefined`. Completely optional.

libs/@guardian/libs/src/consent-management-platform/vendorDataManager.ts
mxdvl: Implemented in #1569

Issue #1058, opened 2024-01-04
`@guardian/source-*` installations


sndrs:
# @guardian/source

Latest version: [`v3.0.0`](https://www.npmjs.com/package/@guardian/source).

### Versions in use

| Installation                                                                                                                                                                                                    |  dependencies | devDependencies | peerDependencies |
| :-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------: | --------------: | ---------------: |
| **workspace:***                                                                                                                                                                                                 |               |                 |                  |
| [csnx/configs/storybook](https://github.com/guardian/csnx/blob/80453631bab88aa3ba5cc60c785011af627a2f81/configs/storybook/package.json)                                                                         | `workspace:*` |                 |                  |
| [dotcom-rendering/apps-rendering](https://github.com/guardian/dotcom-rendering/blob/1a6f86b043639493f06421be3b06e2fc754de591/apps-rendering/package.json)                                                       | `workspace:*` |                 |                  |
| [dotcom-rendering/dotcom-rendering](https://github.com/guardian/dotcom-rendering/blob/1a6f86b043639493f06421be3b06e2fc754de591/dotcom-rendering/package.json)                                                   | `workspace:*` |                 |                  |
| **3**                                                                                                                                                                                                           |               |                 |                  |
| [csnx/libs/@guardian/source-development-kitchen](https://github.com/guardian/csnx/blob/80453631bab88aa3ba5cc60c785011af627a2f81/libs/@guardian/source-development-kitchen/package.json)                         |               |         `3.0.0` |         `^3.0.0` |
| [gateway](https://github.com/guardian/gateway/blob/0f1fff860477d493bd8896a5f202911bc595914d/package.json)                                                                                                       |      `^3.0.0` |                 |                  |
| **2**                                                                                                                                                                                                           |               |                 |                  |
| [support-dotcom-components/packages/modules](https://github.com/guardian/support-dotcom-components/blob/c9b12a6ba75cbfa0b1e511005d978bf2d212f819/packages/modules/package.json)                                 |       `2.0.0` |                 |                  |
| **1**                                                                                                                                                                                                           |               |                 |                  |
| [braze-components](https://github.com/guardian/braze-components/blob/cdb6944faa1d6bfc33051b9f9d2641e06f7a2465/package.json)                                                                                     |               |        `^1.0.1` |         `^1.0.1` |
[...]

PR #1566, opened 2024-06-19
fix source tracker

sndrs:
## What are you changing?

- fixing source tracker

## Why?

- so we can see what versions of source are in use


SiAdcock:
@sndrs  Does this update the spreadsheet as well as [the issue with all the versions](https://github.com/guardian/csnx/issues/1058)?

sndrs:
superseded by #1578

PR #1569, opened 2024-06-20
Simplify `removeData`

mxdvl:
## What are you changing?

- ensure that undefined values default to an empty array so we do not need to check if they are defined.

## Why?

- simpler code, less variations to handle
- less code sent over the wire


Jakeii (approved):

jamesgorrie (approved):
I'm unfamiliar with this code, and it is now easier to read!

sndrs (approved):

PR #1573, opened 2024-06-20
hold off devkit bumps

sndrs:
## What are you changing?

holding off bumping dev kitchn till new source is out

## Why?

otherwise we can't update the lockfile

_this needs fixing, it feels like it should not be this cumbersome!_


oliverabrahams:
> ## What are you changing?
> holding off bumping dev kitchn till new source is out
> 
> ## Why?
> otherwise we can't update the lockfile
> 
> _this needs fixing, it feels like it should not be this cumbersome!_

I agree! Was going to suggest that we remove the need to have the npm version thing it is so tricky to release these packages now

oliverabrahams (approved):

PR #1583, opened 2024-06-25
Remove timeline from format

abeddow91:
## What are you changing?
Remove timeline from Format. 

## Why?
Timelines are now a distinct element, rather than a format design. As such, we would like to remove the timeline design to help enforce usage of the new element and reduce confusion in the codebase.

### DIFF
--- a/.changeset/spicy-dolphins-lay.md
+++ b/.changeset/spicy-dolphins-lay.md
@@ -2,4 +2,6 @@
 '@guardian/libs': major
 ---
 
-Remove timeline from the format
+Removes `Timeline` from `ArticleFormat.design`.
+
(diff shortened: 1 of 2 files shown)


.changeset/spicy-dolphins-lay.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/libs': major
+---
+
+Remove timeline from the format
```
sndrs, line 5:
```suggestion
Removes `Timeline` from `ArticleFormat.design`.

Timelines are now distinct elements, rather than a part of `ArticleFormat`. 
```

abeddow91:
closing as this is not prioritised

PR #1584, opened 2024-06-25
Add web vital attribution data

abeddow91:
## What are you changing?
Updates web-vitals to 

1. Send attribution data with CLS, LCP and INP
2. A new endpoint. The data will now go to the [events collector](https://github.com/guardian/events-collector). The events collector does not have separate endpoints for CODE and PROD environments. Instead, this stage is determined by the isDev parameter and added to the payload to allow us to separate the dataset by stage in big query. 

## Why?
Additional attribution data will help with diagnosing web vital issues on DCR. As well as this, the events collector provides a more agnostic schema with a focus on anonymity of data and is a more fitting pipeline for this dataset.

### DIFF
--- a/.changeset/brave-dryers-fix.md
+++ b/.changeset/brave-dryers-fix.md
@@ -1,0 +1,5 @@
+---
+'@guardian/core-web-vitals': major
+---
+
+Update endpoint and send attribution data to big query
--- a/.changeset/brave-dryers-fix.md
+++ b/.changeset/brave-dryers-fix.md
@@ -1,5 +1,0 @@
----
-'@guardian/core-web-vitals': major
----
-
-Update endpoint and send attribution data to big query
--- a/libs/@guardian/core-web-vitals/package.json
+++ b/libs/@guardian/core-web-vitals/package.json
@@ -33,14 +33,14 @@
 		"ts-jest": "29.1.1",
 		"tslib": "2.6.2",
 		"typescript": "5.3.3",
-		"web-vitals": "3.5.0",
+		"web-vitals": "4.2.0",
 		"wireit": "0.14.4"
 	},
 	"peerDependencies": {
 		"@guardian/libs": "^16.0.0",
 		"tslib": "^2.6.2",
 		"typescript": "~5.3.3",
-		"web-vitals": "^3.5.0"
+		"web-vitals": "^4.2.0"
 	},
 	"peerDependenciesMeta": {
 		"typescript": {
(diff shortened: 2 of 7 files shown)


.changeset/brave-dryers-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/core-web-vitals': major
+---
+
+Update endpoint and send attribution data to big query
```
mxdvl, line 5: Would it be more appropriate to say we added attribution ans the data will end in a different table?

mxdvl (dismissed):
Let’s release a canary of this before merging to ensure it works as expected!

sndrs:
as this is a major, do you want to incorporate #1581?

abeddow91:
> as this is a major, do you want to incorporate #1581?

I think this seems like a good idea, I'll take a look at bumping this package.

libs/@guardian/core-web-vitals/package.json
```diff
@@ -33,7 +33,7 @@
 		"ts-jest": "29.1.1",
 		"tslib": "2.6.2",
 		"typescript": "5.3.3",
-		"web-vitals": "3.5.0",
+		"web-vitals": "4.2.0",
```
sndrs, line 36: don't forget to bump the peerdep version too on line 43 (github won't left me suggest it, sorry)

abeddow91: thanks @sndrs

sndrs: we need to add a linting rule/check for this, sorry you got caught

sndrs (approved):
great from infra side of things. thanks for bumping `web-vitals` too.

.changeset/gorgeous-spiders-applaud.md
```diff
@@ -0,0 +1,8 @@
+---
+'@guardian/core-web-vitals': major
+'@guardian/libs': major
```
sndrs, line 3: sorry, does this need to be in here?

sndrs (changes requested):
i think the libs bump snuck in?

(sorry for the red review, i coudln't work out how to undo the green one and only spotted the libs bump after it!)

.changeset/gorgeous-spiders-applaud.md
abeddow91: 🤦 let me revise that!

sndrs (approved):

pnpm-lock.yaml
abeddow91, line 1: @sndrs - I've removed the changeset for libs but I'm still seeing a canary release for it, I'm guessing because of this change to the lock file? Does this seem usual?

PR #1615, opened 2024-07-23
Add 15px variations for headline typography presets

cemms1:
## What are you changing?

Adds the following to the typography presets:
 - `headlineBold15`
 - `headlineLight15`
 - `headlineLightItalic15`
 - `headlineMedium15`
 - `headlineMediumItalic15`

## Why?

- These styles already exist in the [design system](https://theguardian.design/2a1e5182b/p/01555f-typography-presets) and are required for some upcoming changes to DCR

### DIFF
--- a/.changeset/quiet-brooms-fail.md
+++ b/.changeset/quiet-brooms-fail.md
@@ -3,4 +3,4 @@
 '@guardian/source': minor
 ---
 
-Adds `headlineBold15`, `headlineLight15`, `headlineMedium15` and `headlineMediumItalic15` to the typography presets
+Adds `headlineBold15`, `headlineLight15`, `headlineLightItalic15`, `headlineMedium15` and `headlineMediumItalic15` to the typography presets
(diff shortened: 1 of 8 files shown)


abeddow91 (approved):

.changeset/cold-ads-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/design-tokens': minor
+---
+
+Adds `headlineBold15`, `headlineMedium15` and `headlineMediumItalic15` to the typography presets
```
oliverabrahams, line 5: You want a release to the `@guardian/source` package here. The code from the design tokens package is used to generate the Typography tokens. It is quite confusing... You need to run `make @guardian/source:build-type-presets` then it should generate the new presets in Source

oliverabrahams: Let me know if you want me to go over in more detail

oliverabrahams: Yea this is definitely our fault! It is not clear that updating the tokens does not update Source. I will update the documentation to make this clearer :)

oliverabrahams: We changed how typography presets are made but didn't update the docs. The reason for not directly pulling it from design-tokens in was because really large objects cannot be tree shaken so if u just use 2 typography presets then you still got the whole object.

oliverabrahams: Yea that looks good!

oliverabrahams (approved):

PR #1622, opened 2024-07-25
Refactor DottedLines to use CSS not SVG

mxdvl:
## What are you changing?

- Use simpler CSS to create the dotted line’s pattern
- Follow-up on https://github.com/guardian/source/pull/1320

## Why?

- using SVG with `id` creates invalid HTML if used twice on a page, and can cause visual bug in the first element with a matching `id` is not displayed. See https://github.com/guardian/dotcom-rendering/pull/11985

### DIFF
--- a/.changeset/tiny-ducks-roll.md
+++ b/.changeset/tiny-ducks-roll.md
@@ -1,0 +1,5 @@
+---
+'@guardian/source-development-kitchen': major
+---
+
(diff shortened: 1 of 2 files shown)


sndrs:
these render differently in the chromatic diff, is that expected?

image

sndrs (approved):

PR #1638, opened 2024-08-08
Move `design-tokens` into `@guardian/source`

sndrs:
## What are you changing?

Removes `@guardian/design-tokens` package and adds the tokens directly to `@guardian/source`.

## Why?

Having it as separate package was preparation for a world that never came, and adds complexity and indirection to the process of editing design tokens.

This also allows us to calculate `foundations` values during build, rather than at the point of consumption.

And _that_ relieves user's devices of this overhead and makes it unnecessary to include the design tokens JSON in consumer's bundles.

A CSTI Mobbing Release© by @sndrs, @oliverabrahams and @jamesmockett.


oliverabrahams (approved):

libs/@guardian/source/src/foundations/size/size.ts
```diff
@@ -104,20 +70,15 @@ export const remHeight = {
  ** `width.ctaMedium`: medium height call to action buttons and links
  ** `width.ctaSmall`: small height call to action buttons and links
  ** `width.ctaXsmall`: xsmall height call to action buttons and links
+ ** `width.inputMedium`: text input fields, radio and checkbox labels
+ ** `width.inputXsmall`: checkables such as checkboxes or radio buttons
+ ** `width.inputMedium`
  ** `width.inputXsmall`: checkables such as checkboxes or radio buttons
```
sndrs, line 76: @oliverabrahams are these correct?

PR #1653, opened 2024-08-19
Cc/update ticker

charleycampbell:
## What are you changing?
Adding in new source ticker component
-

## Why?

-

### DIFF
--- a/libs/@guardian/source-development-kitchen/src/react-components/ticker/Ticker.tsx
+++ b/libs/@guardian/source-development-kitchen/src/react-components/ticker/Ticker.tsx
@@ -14,10 +14,19 @@ export interface TickerData {
 	goal: number;
 }
 
+export interface TickerStylingSettings {
+	headlineColour: string;
+	totalColour: string;
+	goalColour: string;
+	filledProgressColour: string;
+	progressBarBackgroundColour: string;
+}
+
 export interface TickerSettings {
 	currencySymbol: string;
 	copy: TickerCopy;
 	tickerData?: TickerData;
+	tickerStylingSettings: TickerStylingSettings;
 }
 
 //styles for the container that holds the ticker
@@ -29,35 +38,38 @@ const tickerContainerStyles = css`
 `;
 
 //styles for headline text (which is optional)
-const headlineStyles = css`
-	${textSansBold17}
+const headlineStyles = (headlineColour: string) => css`
+	${textSansBold17};
+	color: ${headlineColour};
 `;
 
 //styles for the numerical count (total raised so far) and the goal text
-const goalLabelStyles = css`
+const goalLabelStyles = (goalContributionsColour: string) => css`
 	${textSans15};
+	color: ${goalContributionsColour};
 `;
 
-const countLabelStyles = css`
+const countLabelStyles = (countColour: string) => css`
 	${textSansBold17};
-	/* stylelint-disable-next-line color-no-hex */
-	color: #5056f5;
+	color: ${countColour};
 `;
 
-//styles for the ticker background
-const tickerBackgroundStyles = css`
+//styles for the progress bar background
+const progressBarBackgroundStyles = (
+	progressBarBackgroundColour: string,
+) => css`
 	height: 10px;
 	align-self: stretch;
 	align-items: center;
 	display: flex;
 	border-radius: 8px;
-	background: rgba(80, 86, 245, 0.35);
+	background: ${progressBarBackgroundColour};
 	overflow-x: hidden;
 	width: 100%;
 	position: relative;
 `;
 
-//styles for the moving progress bar
+//styles for the moving part of the progress bar
 const progressBarStyles = css`
 	overflow: hidden;
 	width: 100%;
@@ -80,6 +92,7 @@ const progressBarTransform = (goal: number, runningTotal: number): string => {
 const filledProgressStyles = (
 	goal: number,
 	runningTotal: number,
+	progressBarColour: string,
 ): SerializedStyles => css`
 	position: absolute;
 	top: 0;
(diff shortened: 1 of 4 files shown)


libs/@guardian/source-development-kitchen/src/react-components/ticker/Ticker.tsx
```diff
@@ -0,0 +1,167 @@
...
+	totalColour: string;
+	goalColour: string;
+	filledProgressColour: string;
+	progressBarBackgroundColour: string;
+}
+
+export interface TickerSettings {
+	currencySymbol: string;
+	copy: TickerCopy;
+	tickerData?: TickerData;
```
tomrf1, line 28: I know this is how it's modeled elsewhere, but I think we should make this field required now. It means we can remove the default 0 values in the component. It doesn't really make sense to render a ticker with no data

charleycampbell: Good point!

.changeset/chatty-ears-hammer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/source-development-kitchen': major
+---
+
+New ticker component
```
sndrs, line 5: As this is new functionality and won't break any existing uses, i think you can use a `minor` here.

```suggestion
---
'@guardian/source-development-kitchen': minor
---

Add a new `ticker` component
```

sndrs (commented):
looks good - I think you'll need to add the component to `index.ts` so it's importable by consumers?

libs/@guardian/source-development-kitchen/src/react-components/ticker/useIsInView.ts
```diff
@@ -0,0 +1,79 @@
...
+				if (entry.isIntersecting) {
+					setIsInView(true);
+				} else if (options.repeat) {
+					setIsInView(false);
+				}
+			},
+			[options.repeat],
+		);
+
+	const intersectionCallback = intersectionObserverCallback;
```
arelra, line 44: Is this line doing anything now? I noticed the DCR version is using a `debounce` option which is not present here:

https://github.com/guardian/dotcom-rendering/blob/5942fea86286277cbe28e9ca3d4dec2c9b3d6524/dotcom-rendering/src/lib/useIsInView.ts#L52

.changeset/chatty-ears-hammer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/source-development-kitchen': minor
```
charleycampbell, line 2: Good point thanks Alex

libs/@guardian/source-development-kitchen/src/react-components/ticker/useIsInView.ts
```diff
@@ -0,0 +1,79 @@
...
+	/** Set the initial HTML Element, if known. */
+	node?: HTMLElement;
+};
+
+/**
+ * Custom hook around the `IntersectionObserver`.
+ *
+ * @returns a tuple containing `[isInView, setNode]`
+ */
+const useIsInView = (
```
arelra, line 20: Did we ever create anywhere for lib React hooks to live @sndrs ?

I can't see any other hooks in csnx but this hook will now have slightly different versions in DCR and Source

sndrs (approved):

libs/@guardian/source-development-kitchen/src/react-components/ticker/useIsInView.ts
sndrs: nothing currently, if it's useful `libs` could be a good place?

libs/@guardian/source-development-kitchen/src/react-components/ticker/Ticker.tsx
```diff
@@ -0,0 +1,167 @@
...
+	tickerData: TickerData;
+	tickerStylingSettings: TickerStylingSettings;
+}
+
+//styles for the container that holds the ticker
+const tickerContainerStyles = css`
+	display: flex;
+	flex-direction: column;
+	align-items: flex-start;
+	padding: 20px 8px 20px 8px;
```
arelra, line 37: Where possible I think it's preferable to use `space` from `@guardian/source/foundations` so that our spacing conforms to the design system and can be updated in one place.

https://github.com/guardian/source/blob/d84c25d67ed51545994f53986eb4487d35c1aab4/packages/%40guardian/source-foundations/src/space/space.ts#L19

Example usage in DCR:

https://github.com/guardian/dotcom-rendering/blob/9212caca630ce238d67307799facf07afe108abf/dotcom-rendering/src/components/Accordion.tsx#L49

libs/@guardian/source-development-kitchen/src/react-components/ticker/useIsInView.ts
tomrf1: We did wonder about bringing across DCR's version of `useIsInView`. There are a couple of complications:
1. we removed the `debounce` feature, to avoid depending on lodash (and because we don't need it here),
2. it's React-specific, and `@guardian/libs` doesn't currently reference React at all

How do you feel about keeping this alternative version of the hook alongside the ticker for now?

libs/@guardian/source-development-kitchen/src/react-components/ticker/Ticker.tsx
charleycampbell: Yeah really good point. I've used `space` where I can although some of the spacing on the design is custom I can't be used like 7px

libs/@guardian/source-development-kitchen/src/react-components/ticker/useIsInView.ts
arelra: I think thats fine for now but would it be possible to add a comment to point to the DCR version please? Something along the lines of:

https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/components/OnwardsUpper.importable.tsx#L23-L28

In terms of how we consolidate, perhaps something we could discuss as a 'client side' group @sndrs ?

Btw we should be able to import `lodash.debounce` as a single dependency which is a relatively small dep.

https://bundlephobia.com/package/lodash.debounce@4.0.8

charleycampbell: Great! I've added a comment - I'll pop one on the DCR version too.

PR #1665, opened 2024-08-27
Updated AUS privacy manager id

akinsola-guardian:
## What are you changing?

- Updated the privacy manager id for Australia. 

## Why?

- A new privacy manager was created in Sourcepoint and this changes reflect that.

### DIFF
--- a/.changeset/short-starfishes-change.md
+++ b/.changeset/short-starfishes-change.md
@@ -1,0 +1,5 @@
+---
+'@guardian/libs': patch
+---
+
+Updating the Australian Privacy Manager id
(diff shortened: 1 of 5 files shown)


AshCorr (approved):

.changeset/short-starfishes-change.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/libs': patch
+---
+
+Updating the Australian Privacy Manager id
```
sndrs, line 5: Remember that this message is for anyone who uses `@guardian/libs`, so it's useful to be super clear about a change and the reasons for it i.e. "do i need to know about this or can i ignore this one?".

Not sure if I've got the gist of it here, but maybe something like this?

```suggestion
Internally, `cmp.init` uses a new id for Sourcepoint consent config in Australia.

Reflects a change to our setup at Sourcepoint.
```

sndrs (approved):

PR #1668, opened 2024-08-28
Use ESLint 9 to lint `libs/**`

sndrs:
## What are you changing?

- updates the repo to the use the ESLint 9 and the new config added in #1658
- updates `@guardian/eslint-config` slightly to re-enable comments linting
- applies some fixes

## Why?

- staying up to date

## Anything else?

yeah - noticed doing this that we don't lint anything outside `libs/**`, so will add that in a further PR

PR #1669, opened 2024-08-28
Move `tools/scripts/**` to `scripts`

sndrs:
## What are you changing?

have one `scripts` dir, instead of `tools/scripts` and `scripts`. also make `scripts` a package in its own right so we can push the deps out of the root

## Why?

spotted in #1668 that we don't lint `scripts`, this makes it more standalone and ready to be linted


oliverabrahams (approved):
Looks good 👍

PR #1675, opened 2024-08-30
Disallow use of `css` in Source components

sndrs:
## What are you changing?

- disallows use of Emotion's `css` props directly on Source components

## Why?

Although right now Source components _are_ Emotion components, and it's been technically possible to use the `css` prop to customise the styling of Source components, it's not an official API. 

The correct way is to use `cssOverrides`.

However, because nothing prevented it, it's found its way into a few places that are now breaking when trying to upgrade to `@emotion/react` v11.11.3 (require for `source` v4).

This change removes that foot gun by preventing further uses.

### DIFF
--- a/.changeset/chatty-steaks-tap.md
+++ b/.changeset/chatty-steaks-tap.md
@@ -1,0 +1,5 @@
+---
+'@guardian/source': major
+---
+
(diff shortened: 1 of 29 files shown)


arelra (approved):
✨

joecowton1 (approved):

PR #1676, opened 2024-08-30
removed overrides because they are no longer needed

oliverabrahams:
Removed overrides 💯 


sndrs (approved):
be gone

PR #1693, opened 2024-09-06
Set internal devDep versions of `libs` to bottom of peerDep range

sndrs:
## What are you changing?

- some devDep versions had drifted from the peerDep min version (e.g. #1692), this restores them
- sets the devDep version of `libs` in `source` to workspace version, because it doesn't have a peerDep 

## Why?

- ensures we develop against the lowest version of peerDeps we declare compatibility with

### DIFF
--- a/libs/@guardian/source/package.json
+++ b/libs/@guardian/source/package.json
@@ -32,8 +32,7 @@
 		"storybook": "storybook dev --port 4401",
 		"test": "wireit",
 		"tsc": "wireit",
-		"verify-dist": "wireit",
-		"wireit": "storybook build"
+		"verify-dist": "wireit"
 	},
 	"dependencies": {
 		"mini-svg-data-uri": "1.4.4"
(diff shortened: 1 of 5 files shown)


libs/@guardian/source/package.json
```diff
@@ -32,8 +32,7 @@
 		"storybook": "storybook dev --port 4401",
 		"test": "wireit",
 		"tsc": "wireit",
-		"verify-dist": "wireit",
-		"wireit": "storybook build"
```
sndrs, line 36: i think this crept in from #1685 and was wrong

libs/@guardian/source/package.json
```diff
@@ -46,7 +45,7 @@
 		"@cobalt-ui/plugin-js": "1.4.3",
 		"@cobalt-ui/utils": "1.2.2",
 		"@emotion/react": "11.11.3",
-		"@guardian/libs": "18.0.2",
```
oliverabrahams, line 49: Nice!!!

oliverabrahams (approved):
Looks great! like the workspace v in source

PR #1706, opened 2024-09-16
Use gpcEnabled and factor into doNotSell flag

akinsola-guardian:
## What are you changing?

-

## Why?

-


akinsola-guardian:
Closed and re-opened here  https://github.com/guardian/csnx/pull/1740 because this is using ccpa.

PR #1719, opened 2024-10-01
Configurable ticker size

tomrf1:
Adds a new `size` prop to the ticker component. The style/font changes depending on this value.

I've also done some renaming:
1. The `countLabel` field is now called `headline`, and is optional.
2. previously "count" and "total" were used interchangeably, now it's only "total"

### DIFF
--- a/libs/@guardian/source-development-kitchen/src/react-components/ticker/Ticker.tsx
+++ b/libs/@guardian/source-development-kitchen/src/react-components/ticker/Ticker.tsx
@@ -49,15 +49,15 @@ const headlineStyles = (headlineColour: string) => css`
 	padding-bottom: ${space[1]}px;
 `;
 
-//styles for the numerical count (total raised so far) and the goal text
-const goalLabelStyles = (goalContributionsColour: string) => css`
+//styles for the total and the goal text
+const goalLabelStyles = (goalColour: string) => css`
 	${textSans15};
-	color: ${goalContributionsColour};
+	color: ${goalColour};
 `;
 
-const countLabelStyles = (countColour: string) => css`
+const totalLabelStyles = (totalColour: string) => css`
 	${textSansBold17};
-	color: ${countColour};
+	color: ${totalColour};
 `;
 
 const labelContainerStyles = css`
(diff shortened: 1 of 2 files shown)


libs/@guardian/source-development-kitchen/src/react-components/ticker/Ticker.tsx
```diff
@@ -143,25 +157,33 @@ export const Ticker = ({
...
 
+	const { barHeight, totalFont } = tickerSizeSettings[size];
+
 	return (
 		<div ref={setNode}>
 			<div>
 				<div css={tickerContainerStyles}>
-					<div css={headlineStyles(headlineColour)}>{copy.countLabel}</div>
-					<div css={progressBarBackgroundStyles(progressBarBackgroundColour)}>
-						<div css={progressBarStyles}>
```
tomrf1, line 152: I noticed this div doesn't do anything, so I've removed it

.changeset/modern-scissors-hide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/source-development-kitchen': major
+---
+
+Make ticker size configurable
```
oliverabrahams, line 5: You could add some of the information about renaming properties and the new size argument

.changeset/modern-scissors-hide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/source-development-kitchen': major
```
sndrs, line 2: if this only adds a `size` option, i think it's a minor?

.changeset/modern-scissors-hide.md
sndrs, line 5:
```suggestion
Make `Ticker` size configurable via a new `size` prop
```

.changeset/modern-scissors-hide.md
tomrf1: I went with major because a) the `size` prop is required, and b) this PR also renames an existing field. Which I guess makes it not backwards compatible

tomrf1: (we're not actually using this component yet)

.changeset/modern-scissors-hide.md
tomrf1: 👍 just pushed a better description

oliverabrahams (approved):

PR #1342, opened 2024-04-08
CMP: Implement the Multi-State-Privacy-Agreement in the US

akinsola-guardian:
## What are you changing?

Transition to GPP API for US National Privacy Framework (usnat)

- Updated to use the __gpp API command for the "usnat" privacy framework.
- Updated usnat privacy manager id to use MSPS message
- Adjusted sourcepoint configuration settings for usnat
- Update US framework from ccpa to usnat 
- Moved CCPA-related logic to the Australian (AUS) namespace.
- Updated the end-to-end (e2e) test for "usnat" to check whether the banner is visible after user interaction, rather than checking the data-donotsell property.

## Why?

- IAB is deprecating US privacy string and moving to the multi-state privacy agreement. 

This has been tested using the github pages locally and tested in dotcom-rendering in CODE.

PR #1720, opened 2024-10-01
Revert "Remove `ArticleDesign.PrintShop`"

SiAdcock:
Reverts guardian/csnx#1683

libs has 2 breaking changes at v19 that we'd like to decouple. The change that removes `ArticleDesign.PrintShop` and a change to the CMP (#1342)

To move quickly on the CMP release in DCR, we'd like to roll back the PrintShop changes and re-implement them at a later date.

### DIFF
--- a/.changeset/small-bats-help.md
+++ b/.changeset/small-bats-help.md
@@ -2,4 +2,4 @@
 '@guardian/libs': minor
 ---
 
-Revert "Remove `ArticleDesign.PrintShop`". To be re-implemented at a later date.
+Restores `ArticleDesign.PrintShop`. To be re-removed at a later date.
--- a/.changeset/small-bats-help.md
+++ b/.changeset/small-bats-help.md
@@ -1,5 +1,5 @@
 ---
-"@fake-scope/fake-pkg": major
+"@fake-scope/fake-pkg": minor
 ---
 
 Revert "Remove `ArticleDesign.PrintShop`". To be re-implemented at a later date.
--- a/.changeset/small-bats-help.md
+++ b/.changeset/small-bats-help.md
@@ -1,5 +1,5 @@
 ---
-"@fake-scope/fake-pkg": minor
+'@fake-scope/fake-pkg': minor
 ---
 
 Revert "Remove `ArticleDesign.PrintShop`". To be re-implemented at a later date.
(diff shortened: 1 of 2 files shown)


.changeset/small-bats-help.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fake-scope/fake-pkg": major
```
sndrs, line 2: i think this would be a minor right? and then we can deprecate 19.0.0?

sndrs, line 2: i think this would be a minor right? and then we can deprecate v19.0.0?

```suggestion
"@fake-scope/fake-pkg": minor
```

.changeset/small-bats-help.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fake-scope/fake-pkg": minor
```
jamesmockett, line 2: Unless I'm missing something this should be `@guardian/libs`?

```suggestion
"@guardian/libs": minor
```

.changeset/small-bats-help.md
```diff
@@ -0,0 +1,5 @@
+---
+'@fake-scope/fake-pkg': minor
```
sndrs, line 2: is this right?
```suggestion
'@fake-scope/fake-pkg': minor
```

sndrs, line 2:
```suggestion
'@guardian/libs': minor
```

sndrs (changes requested):

.changeset/small-bats-help.md
SiAdcock: Sorry, I charged ahead with the automated changeset without looking closely enough 🙈

sndrs (approved):

.changeset/small-bats-help.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/libs': minor
+---
+
+Revert "Remove `ArticleDesign.PrintShop`". To be re-implemented at a later date.
```
sndrs, line 5:
```suggestion
Restores `ArticleDesign.PrintShop`. To be re-removed at a later date.
```

sndrs (approved):

PR #1732, opened 2024-10-09
Remove the unused format prop from ToggleSwitch

SiAdcock:
## What are you changing?

- Remove the unused `format` prop from the `ToggleSwitch` component in `@guardian/source-development-kitchen`

## Why?

The optional format prop is never used anywhere in the Guardian organisation, implying it is redundant. 

The component doesn't really do a lot with the format. If a `format` is set, the element gets a slightly different colour palette, but we aren't checking the value of the prop `format`, just whether or not it exists. At best this is overly complex. 

**Without format:**

Screenshot 2024-10-09 at 12 32 52

**With format:** 

Screenshot 2024-10-09 at 12 32 43

Note: background colour doesn't get applied automatically, this must be manually added to the containing element

If we decide we want to keep the alternative colours, we can do so using a flag rather than a `format` prop e.g. `altColourScheme: boolean`. But it should be noted that this alternative colour scheme [is not reflected in the Design System](https://www.figma.com/design/FnCwHWaXPUsyjE0luiI03e/%E2%97%89-Cheat-sheet?m=auto&node-id=2437-73573&t=YGnpuBQPWo6cJ8rR-1).

### DIFF
--- a/.changeset/tame-geese-bathe.md
+++ b/.changeset/tame-geese-bathe.md
@@ -1,0 +1,5 @@
+---
+'@guardian/source-development-kitchen': major
+---
+
+Remove format prop from Toggle Switch component
--- a/.changeset/tame-geese-bathe.md
+++ b/.changeset/tame-geese-bathe.md
@@ -2,4 +2,6 @@
 '@guardian/source-development-kitchen': major
 ---
 
-Remove format prop from Toggle Switch component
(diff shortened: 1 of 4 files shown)


domlander (approved):

sndrs (approved):
i love this! was going to propose an alternative prop too, but this is miles better. 

:shipit:

PR #1733, opened 2024-10-09
Make Select show error styling only if user has interacted

andrewHEguardian:
## What are you changing?

- Make the `Select` element only use the error styling if it is [user-invalid](https://developer.mozilla.org/en-US/docs/Web/CSS/:user-invalid) (i.e. the user has interacted with it and it fails validation)

## Why?

- If the `Select` element had the `required` attribute, it would be styled as an error before a user has even interacted with the form, as in the screenshot below.

image


GHaberis:
@andrewHEguardian does this mean if a user were to scroll past a required select field and hit submit the error state would not be shown because they didn't interact with the field?

andrewHEguardian:
> @andrewHEguardian does this mean if a user were to scroll past a required select field and hit submit the error state would not be shown because they didn't interact with the field?

An interaction with the form (ie submitting it) would also trigger the validation. 

See this [example](https://developer.mozilla.org/en-US/play?id=qb9bJD2pgH1mjfTX7mk8I8GdztlCsritrqlqJqLsW6uCkcvF9JQ0aaKrdW2JP0jQsgkc8OfEAstZQSsT)

GHaberis:
> > @andrewHEguardian does this mean if a user were to scroll past a required select field and hit submit the error state would not be shown because they didn't interact with the field?
> 
> An interaction with the form (ie submitting it) would also trigger the validation.
> 
> See this [example](https://developer.mozilla.org/en-US/play?id=qb9bJD2pgH1mjfTX7mk8I8GdztlCsritrqlqJqLsW6uCkcvF9JQ0aaKrdW2JP0jQsgkc8OfEAstZQSsT)

Excellent, thanks for clarifying and the example!

GHaberis (approved):
Looks good to me 👍

rupertbates (commented):
From @sndrs - :user-invalid only has 87% support at the moment, does it matter that this won't fix anything for 13% of users?

andrewHEguardian:
> :user-invalid only has 87% support at the moment, does it matter that this won't fix anything for 13% of users?

It would mean users in that 13% wouldn't see the red border unless the `error` prop was passed. 

I'm not sure this matters too much, but if we want to be careful we could wrap it in a `@supports` rule?

SiAdcock:
Thanks @andrewHEguardian, this is a good change 👍 

I'd recommend using the [`optional` prop](https://theguardian.design/2a1e5182b/p/663879-select-box/t/page-663879-38205517-14) rather than the HTML `required` attribute as it follows Source conventions such as automatically adding the "Optional" hint to the label.

In future we should revisit the [controlled vs uncontrolled](https://github.com/guardian/source/issues/360) question, as maintaining a library that allows both control mechanisms adds maintenance burden. This is beyond the scope of this PR.

SiAdcock (approved):

sndrs:
while this is definite improvement for people who will experience the new behaviour, i think this means that it's now possible that if you only use the `required` attribute for validation (i.e. no internal component validation), the 13% of users who currently do not support the `:user-invalid` pseudo selector will not get visual feedback indicating an invalid state. 

we need these components to offer a good experience to as many people as possible, and for those people this means a loss of functionality.

@oliverabrahams and I are looking into ways support this change without affecting people who cannot benefit from it, so for now we've added a "Do not merge" label to the release PR

sndrs:
we opted to revert this to unblock releases from the repo. once we have a way to achieve your change with wider support we'll recreate this PR - sorry @andrewHEguardian!

PR #1757, opened 2024-11-01
Make `Select` display invalid state only if user has interacted, take 2

sndrs:
Restores and provides a fallback for @andrewHEguardian's original change from #1733.

If this is approved we can copy the solution to the rest of the input components.


andrewHEguardian (approved):
Great!

PR #1762, opened 2024-11-05
Add `react-crossword` package to workspace

sndrs:
## What are you changing?

Adds a new, v3 of [`@guardian/react-crossword`](https://github.com/guardian/react-crossword).

It's very much in an alpha state, and none of this will be published from here, from now.

But it sets up a probably new home for the NPM package.

## Why?

As part of the migration to DCR, we came across this package. We plan to coalesce our work into that package as a new major version.

Co-authored-by: @oliverabrahams


oliverabrahams (approved):
Nice!

PR #1765, opened 2024-11-06
Always add `required` attribute for non-optional `Select`s

sndrs:
## What are you changing?

> [!IMPORTANT]
> requires #1757

- `<Select optional={false} />` will now automatically set `required` on the internal `<select>` element
- `Select` no longer accepts a `required` prop 

## Why?

If a server-rendered `Select` is required we want the browser to enforce that if the user interacts with the native `form` before the component has mounted/hydrated/taken over.

Currently, is also possible to set `optional` and `required` to both be true:

```tsx
<Select optional={true} required={true} />
```

and to pass `success` to an invalid component:

```tsx
<Select success="Well done" required={true} /> // nothing selected
```

which results in the following: 

image

This makes that impossible.

As with #1757, if this is approved, we can copy the behaviour to the other form components.


andrewHEguardian (approved):
👍

SiAdcock (approved):
Let's do this 👍

PR #1769, opened 2024-11-11
focus and progress state in crossword component

sndrs:
## What are you changing?

some proposed refactors that i found helped me to understand the code. see what you think.

- added `cellSize` to `theme`
- broke up `focus` state into `currentEntryId` and `currentCell`
- added col/row id to the svg cells, so we don't have to calculate cells coords from click location
- the rest is mainly variable names or code order i think, without changing how it works

## Why?

`currentEntryId` or `currentCell` could change without the other needing to, but anything that relied on `focus` would have to re-render e.g. the highlighted clue in the clue list


oliverabrahams (approved):
Great

PR #1773, opened 2024-11-11
Allow arrow movement over blank squares

oliverabrahams:
## What are you changing?

- Allows movement over blank squares - same as the APH crossword which is made with accessibility in mind. Current thoughts are this will be better for accessibility. - https://crossword.aphtech.org/
- if you press an arrow key it moves one square

## Why?

- Easier to navigate the whole grid if you cannot see which squares are blank


libs/@guardian/react-crossword/src/components/Crossword.tsx
```diff
@@ -40,7 +40,7 @@ export const Crossword = ({ theme: userTheme, ...props }: CrosswordProps) => {
 	});
 
 	const moveFocus = useCallback(
-		(delta: { x: number; y: number }) => {
+		(delta: { x: number; y: number }, typing = false) => {
```
sndrs, line 43: if you make this an object, it helps describe what that `true` is doing further down:

`moveFocus({ x: -1, y: 0 }, true);` → `moveFocus({ x: -1, y: 0 }, { typing: true });`

```suggestion
		(delta: { x: number; y: number }, { typing = false }: { typing: boolean }) => {
```

sndrs (approved):
one suggestion, but looks great

libs/@guardian/react-crossword/src/components/Crossword.tsx
oliverabrahams: Thanks, I changed it to a single object (delta: {}, isTyping = false: {delta: {}; isTyping: boolean})

PR #1776, opened 2024-11-11
add local storage to Crossword

oliverabrahams:
## What are you changing?

- Saves the progress in local storage
- Retrieves progress in local storage on initialisation
- Retrieves progress on 'storage' event https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event

## Why?

- The progress for the crossword needs to be saved.


sndrs (approved):
reviewed offline

PR #1781, opened 2024-11-13
Grey out completed clues

sndrs:
## What are you changing?

- greying out completed clues

## Why?

- progress cannot be stopped

### DIFF
--- a/libs/@guardian/react-crossword/src/components/Clues.tsx
+++ b/libs/@guardian/react-crossword/src/components/Clues.tsx
@@ -1,6 +1,10 @@
 import { css } from '@emotion/react';
-import type { CAPIEntry } from '../@types/CAPI';
-import type { CurrentEntryId, Progress, Theme } from '../@types/crossword';
+import type {
+	CurrentEntryId,
+	Entries,
+	Progress,
+	Theme,
+} from '../@types/crossword';
 import type { Direction } from '../@types/Direction';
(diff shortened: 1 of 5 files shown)


libs/@guardian/react-crossword/src/components/Clues.tsx
```diff
@@ -1,9 +1,6 @@
 import { css } from '@emotion/react';
-import type {
-	CrosswordEntry,
-	CurrentEntryId,
-	Theme,
-} from '../@types/crossword';
+import type { CAPIEntry } from '../@types/CAPI';
```
sndrs, line 2: this feels like a strange change to need to make - maybe we need to check the types

libs/@guardian/react-crossword/src/components/Clues.tsx
```diff
@@ -13,12 +10,19 @@ const title = css`
 
 type Props = {
 	direction: Direction;
-	entries: CrosswordEntry[];
+	entries: CAPIEntry[];
```
sndrs, line 13: this feels like a strange change to need to make - maybe we need to check the types

oliverabrahams: That function could return a Map<EntryId, CrosswordEntry>  and then Array.from(entries.values())...

oliverabrahams (approved):

PR #1786, opened 2024-11-14
Add separators to crossword

sndrs:
## What are you changing?

- Adds separators to crossword

## Why?

- we need them

image


oliverabrahams (approved):
100/100

PR #1790, opened 2024-11-15
Handle grouped entries

sndrs:
## What are you changing?

hightlight clues/cells of entries that are grouped

image


oliverabrahams (approved):
Looks Good

libs/@guardian/react-crossword/src/components/Cell.tsx
```diff
@@ -19,13 +23,14 @@ export const CellComponent = ({
 	guess = '',
 	theme,
 	isFocused,
-	isHighlighted,
+	isRelated,
+	isActive,
```
oliverabrahams, line 27: Could we make this consistent with the clues? isHighlighted / isActive

sndrs: yeah 100% - i'm sure i've done that somewhere else, will do it here though

PR #1798, opened 2024-11-18
Assume any separator that’s not a hyphen is a thicker line

sndrs:
## What are you changing?

treats any separator that is not `-` as a thicker border

image

## Why?

@andrew-nowak pointed out that we have some that use `/`, `.` etc that are not currently handled


andrew-nowak (approved):
lgtm! thanks!

oliverabrahams (approved):

PR #1802, opened 2024-11-20
Use `use-local-storage-state` for crosswords

sndrs:
## What are you changing?

- replaces our own state-to-`localStorage` handling with [`use-local-storage-state`](https://www.npmjs.com/package/use-local-storage-state)
- renames `updateProgress` → `setCellProgress` (i was getting update/set mixed up when reading)
- updated `revealGrid` to make one change to the state (rather than one call per cell, which was causing a lot of local storage writes)

## Why?

Simpler code for us, and better performance for users


oliverabrahams (approved):
Nice

PR #1804, opened 2024-11-21
Fix for usnat supported api lookup

arelra:
## Why?

Since 19/11/2034 ~ 18:00 UTC the consent state property `doNotSell` is being set to true for all US users regardless of their consent choice. This mean third party integrations that check for `doNotSell` before loading are not running e.g. prebid, comscore and a9 (Amazon ads).

### gppData

We read the `gppData` object as provided by Sourcepoint:

### supportedAPIs string

In order to lookup the usnat metadata we read the `gppData.supportedAPIs` property and split the string to set the `supportedAPIs` string (e.g. `usnat`) 

### supportedAPIs lookup

We then use the `supportedAPIs` string to lookup the property on `gppData.parsedSection[supportedAPIs]`

However currently the `supportedAPIs` string is resolving to `usnat` but the property on `gppData.parsedSection` is `usnatv1`

This causes the lookup to fail and sets `doNotSell` to true.

## What are you changing?

This change updates the logic so rather than using the `supportedAPIs` string it directly attempts to read the first object in `gppData.parsedSection`.

### DIFF
--- a/libs/@guardian/libs/src/consent-management-platform/usnat/getConsentState.ts
+++ b/libs/@guardian/libs/src/consent-management-platform/usnat/getConsentState.ts
@@ -1,18 +1,29 @@
 import { type USNATConsentState } from '../types/usnat';
 import { getGPPData } from './api';
 
+const getSupportedAPIv1 = (
+	supportedAPI: string | undefined,
+): string | undefined => {
+	if (supportedAPI?.includes('usnat')) {
+		return supportedAPI.includes('v1') ? supportedAPI : `${supportedAPI}'v1'`;
+	}
+	return supportedAPI;
+};
+
 export const getConsentState: () => Promise<USNATConsentState> = async () => {
 	let doNotSell = false; // Opt-Out
 	const gppData = await getGPPData();
 
-	const supportedAPIs = gppData.supportedAPIs[0]?.split(':')[1]; // E.G: '7:usnatv1', '8:uscav1'
+	const supportedAPI = gppData.supportedAPIs[0]?.split(':')[1]; // E.G: '7:usnatv1', '8:uscav1'
+	// Temporary fix 21/11/2024
+	const supportedAPIv1 = getSupportedAPIv1(supportedAPI);
 
-	if (supportedAPIs) {
-		//0 Not Applicable. SharingOptOutNotice value was not applicable or no notice was provided, 1 Opted Out, 2 Did Not Opt Out
-		doNotSell =
-			gppData.parsedSections[supportedAPIs]?.SaleOptOut !== 2 ||
-			gppData.parsedSections[supportedAPIs].Gpc;
+	if (supportedAPIv1) {
 		// https://github.com/InteractiveAdvertisingBureau/Global-Privacy-Platform/blob/main/Sections/US-National/IAB%20Privacy%E2%80%99s%20National%20Privacy%20Technical%20Specification.md
+		// 0 Not Applicable. SharingOptOutNotice value was not applicable or no notice was provided, 1 Opted Out, 2 Did Not Opt Out
+		doNotSell =
+			gppData.parsedSections[supportedAPIv1]?.SaleOptOut !== 2 ||
+			gppData.parsedSections[supportedAPIv1].Gpc;
 	}
 
 	return {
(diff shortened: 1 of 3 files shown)


libs/@guardian/libs/src/consent-management-platform/usnat/getConsentState.ts
```diff
@@ -1,18 +1,29 @@
 import { type USNATConsentState } from '../types/usnat';
 import { getGPPData } from './api';
 
+const getSupportedAPIv1 = (
+	supportedAPI: string | undefined,
+): string | undefined => {
+	if (supportedAPI?.includes('usnat')) {
+		return supportedAPI.includes('v1') ? supportedAPI : `${supportedAPI}'v1'`;
```
Jakeii, line 8: I think you've got some extra quotes here?
```suggestion
		return supportedAPI.includes('v1') ? supportedAPI : `${supportedAPI}v1`;
```

arelra: Thanks!

libs/@guardian/libs/src/consent-management-platform/usnat/getConsentState.ts
```diff
@@ -1,18 +1,29 @@
...
-	if (supportedAPIs) {
-		//0 Not Applicable. SharingOptOutNotice value was not applicable or no notice was provided, 1 Opted Out, 2 Did Not Opt Out
-		doNotSell =
-			gppData.parsedSections[supportedAPIs]?.SaleOptOut !== 2 ||
-			gppData.parsedSections[supportedAPIs].Gpc;
+	if (supportedAPIv1) {
 		// https://github.com/InteractiveAdvertisingBureau/Global-Privacy-Platform/blob/main/Sections/US-National/IAB%20Privacy%E2%80%99s%20National%20Privacy%20Technical%20Specification.md
+		// 0 Not Applicable. SharingOptOutNotice value was not applicable or no notice was provided, 1 Opted Out, 2 Did Not Opt Out
+		doNotSell =
+			gppData.parsedSections[supportedAPIv1]?.SaleOptOut !== 2 ||
```
Jakeii, line 25: is there any issue with doing `gppData.parsedSections['usnatv1']?.SaleOptOut` instead?

sndrs (approved):

libs/@guardian/libs/src/consent-management-platform/usnat/getConsentState.ts
```diff
@@ -1,18 +1,29 @@
 import { type USNATConsentState } from '../types/usnat';
 import { getGPPData } from './api';
 
+const getSupportedAPIv1 = (
+	supportedAPI: string | undefined,
+): string | undefined => {
+	if (supportedAPI?.includes('usnat')) {
```
akinsola-guardian, line 7: This will for the majority of states but there are 5 states that have their own id's.   'usnat', 'usca','usva', 'usco', 'usut', 'usct'

libs/@guardian/libs/src/consent-management-platform/usnat/getConsentState.ts
arelra: I think we saw when we were debugging that this code ran for all the frameworks so didn't want to hardcode to `usnat`.

akinsola-guardian: There are 5 states that have their own id's/frameworks. 'usnat', 'usca','usva', 'usco', 'usut', 'usct'

akinsola-guardian: hence, we need to check the supportedAPis to check the framework. There are other states that might follow these 5 in the future so it has be dynamically checked.

libs/@guardian/libs/src/consent-management-platform/usnat/getConsentState.ts
arelra: Thanks @akinsola-guardian have updated

Jakeii (approved):

arelra:
@akinsola-guardian Thanks for letting us know Sourcepoint have fixed this from their side. Will leave this PR for your consideration. If we can depend on always reading the first entry in `gppData.parsedSection` the updated logic should be more resilient to key name mismatches to `gppData.supportedAPIs`

sndrs:
just a reminder that we disabled the CMP e2e tests in #1803 so they won't be running on this PR...

ashishpuliyel (approved):

PR #1806, opened 2024-11-21
Say that crosswords are saved

sndrs:
_depends on #1805_

## What are you changing?

displays a message saying that crosswords are saved (or that they're not, if not)

### `localStorage` is available
image

### `localStorage` is not available
image

## Why?

parity with prod, and a little bit more


oliverabrahams (approved):
Cool

PR #1808, opened 2024-11-22
Zetterise crossword input

sndrs:
## What are you changing?

handle entries with numbers and diacritics in their solution

## Why?

https://github.com/guardian/frontend/pull/20733 – always in our hearts

### DIFF
--- a/libs/@guardian/react-crossword/src/components/Crossword.tsx
+++ b/libs/@guardian/react-crossword/src/components/Crossword.tsx
@@ -161,9 +161,8 @@ export const Crossword = ({
 					break;
 				}
 				default: {
-					const upperCaseKey = key.toUpperCase();
-					if (keyDownRegex.test(upperCaseKey) && currentEntryId) {
-						setCellProgress({ ...currentCell, value: upperCaseKey });
+					if (currentEntryId && keyDownRegex.test(key)) {
+						setCellProgress({ ...currentCell, value: key.toUpperCase() });
 						if (direction === 'across') {
 							moveFocus({ delta: { x: 1, y: 0 }, isTyping: true });
 						}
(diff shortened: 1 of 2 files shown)


libs/@guardian/react-crossword/src/components/Crossword.tsx
```diff
@@ -15,6 +15,9 @@ import { Clues } from './Clues';
 import { Controls } from './Controls';
 import { Grid } from './Grid';
 
+// define and cache the regex for valid keydown events
+const keyDownRegex = /^[A-Za-zÀ-ÿ0-9]$/;
```
oliverabrahams, line 19: do we want lowercase a-z ?

sndrs: that's a good point. if we've already `uppercase`d it there's no need (and it's also what's happening [in frontend](https://github.com/guardian/frontend/blob/main/static/src/javascripts/projects/common/modules/crosswords/crossword.js#L361)). 

I've slightly rejigged the logic to test against all valid keyboard inputs, and then store a valid input as uppercased if it passes.

libs/@guardian/react-crossword/src/components/Crossword.tsx
```diff
@@ -158,9 +161,8 @@ export const Crossword = ({
 					break;
 				}
 				default: {
-					const upperCaseKey = key.toUpperCase();
-					if (/^[A-Z]$/.test(upperCaseKey) && currentEntryId) {
-						setCellProgress({ ...currentCell, value: upperCaseKey });
+					if (currentEntryId && keyDownRegex.test(key)) {
+						setCellProgress({ ...currentCell, value: key.toUpperCase() });
```
sndrs, line 165: trying to make the conditional fail as fast as possible and only perform actions like uppercasing if we need to

oliverabrahams (approved):

PR #1810, opened 2024-11-22
Make crossword theme settings standard props

sndrs:
## What are you changing?

- replaces the `theme` prop with the different theme options as top-level props of `Crossword`
- adds a user-themed story
  image

## Why?

it's more reactish


oliverabrahams (approved):
💯

PR #1811, opened 2024-11-22
Prop-drill `progress`

sndrs:
## What are you changing?

prop-drilling `progress` instead of using `context`

## Why?

the use of context for progress in #1791 is overkill. this is all we need.


libs/@guardian/react-crossword/src/hooks/useProgress.ts
```diff
@@ -109,3 +109,5 @@ export const useProgress = (data: CAPICrossword, userProgress?: Progress) => {
 		isPersistent,
 	] as const;
 };
+
+export type UseProgress = ReturnType<typeof useProgress>;
```
oliverabrahams, line 113: How do you feel about making this an object. I only say that because the type UseProgress[1] seems less descriptive than UseProgress['setProgress']. I see that this is more similar to useState etc.

oliverabrahams (approved):

libs/@guardian/react-crossword/src/hooks/useProgress.ts
sndrs: i had that same thought yeah, i think it's a good idea

sndrs: done in 49e1718

PR #1825, opened 2024-11-29
Single interactive grid component

oliverabrahams:
## What are you changing?

- Move the controls and grid into a single component for layout purposes

## Why?

- as the grid and controls are linked and props will be shared, it will be simpler to have them as a single layout component


libs/@guardian/react-crossword/src/components/Layout.tsx
```diff
@@ -2,28 +2,11 @@ import { css } from '@emotion/react';
...
-};
-
-const useWidthForCols = (cols: number) => {
-	const gridWidth = useGridWidth();
-	const { clueMinWidth } = useTheme();
-
-	return gridWidth + clueMinWidth * cols + 'px';
-};
+import { useGridWidth } from '../hooks/useGridWidth';
+import { useWidthForCols } from '../hooks/useWidthForCols';
```
sndrs, line 9: are these needed elsewhere now?

oliverabrahams: Not at the moment but i they were at one point in this branch and i liked having them separate

sndrs (approved):

PR #1834, opened 2024-12-04
Refactor custom layout

sndrs:
## What are you changing?

- refactors custom layouts into a `Layout` props of `Crossword`
- moves UI state into context
- breaks up `InteractiveGrid` again (so you can handle `Controls` separately after all) 

## Why?

preparation for adding a print layout


oliverabrahams (approved):
Looks good

PR #1846, opened 2024-12-10
Create const for sourcepoint choice type id

akinsola-guardian:
## What are you changing?

- Creating const for choice type id's

## Why?

- Clearer to understand the different choice type id's.

### DIFF
--- a/libs/@guardian/libs/src/consent-management-platform/lib/sourcepointConfig.ts
+++ b/libs/@guardian/libs/src/consent-management-platform/lib/sourcepointConfig.ts
@@ -13,8 +13,8 @@ export const ENDPOINT = isGuardianDomain()
 export type EndPoint = typeof ENDPOINT;
 
 // https://docs.sourcepoint.com/hc/en-us/articles/4405397484307-Event-callbacks-CMP#h_01FTY32EGZ7H3SS192MEE6SNCP
-export enum SourcePointChoiceTypes {
-	AcceptAll = 11,
-	RejectAll = 13,
-	Dismiss = 15,
-}
+export const SourcePointChoiceTypes = {
+	AcceptAll: 11,
+	RejectAll: 13,
+	Dismiss: 15,
+} as const;
--- a/libs/@guardian/libs/src/consent-management-platform/lib/sourcepointConfig.ts
+++ b/libs/@guardian/libs/src/consent-management-platform/lib/sourcepointConfig.ts
@@ -11,3 +11,10 @@ export const ENDPOINT = isGuardianDomain()
 	? 'https://sourcepoint.theguardian.com'
 	: 'https://cdn.privacy-mgmt.com';
 export type EndPoint = typeof ENDPOINT;
+
+// https://docs.sourcepoint.com/hc/en-us/articles/4405397484307-Event-callbacks-CMP#h_01FTY32EGZ7H3SS192MEE6SNCP
+export enum SourcePointChoiceTypes {
+	AcceptAll = 11,
+	RejectAll = 13,
+	Dismiss = 15,
+}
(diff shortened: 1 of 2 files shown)


libs/@guardian/libs/src/consent-management-platform/sourcepoint.ts
```diff
@@ -134,11 +135,11 @@ export const init = (framework: ConsentFramework, pubData = {}): void => {
...
 					log('cmp', `onMessageChoiceSelect choice_type_id: ${choiceTypeID}`);
+					// https://documentation.sourcepoint.com/web-implementation/web-implementation/multi-campaign-web-implementation/event-callbacks#choice-type-id-descriptions
 					if (
-						// https://documentation.sourcepoint.com/web-implementation/web-implementation/multi-campaign-web-implementation/event-callbacks#choice-type-id-descriptions
-						choiceTypeID === 11 ||
-						choiceTypeID === 13 ||
-						choiceTypeID === 15
+						Object.values(SourcePointChoiceTypes).some(
+							(spChoiceType) => spChoiceType === choiceTypeID,
+						)
```
pvighi, line 142: I don't really have a ton ts experience, but wouldn't something like this work ?
```typescript
if ( choiceTypeID  in SourcePointChoiceTypes) {
...
}
```

libs/@guardian/libs/src/consent-management-platform/lib/sourcepointConfig.ts
```diff
@@ -11,3 +11,10 @@ export const ENDPOINT = isGuardianDomain()
 	? 'https://sourcepoint.theguardian.com'
 	: 'https://cdn.privacy-mgmt.com';
 export type EndPoint = typeof ENDPOINT;
+
+// https://docs.sourcepoint.com/hc/en-us/articles/4405397484307-Event-callbacks-CMP#h_01FTY32EGZ7H3SS192MEE6SNCP
+export enum SourcePointChoiceTypes {
+	AcceptAll = 11,
+	RejectAll = 13,
+	Dismiss = 15,
+}
```
coldlink, line 20: We generally want to avoid using enums due to the way they're represented once compiled can cause complications on the way we think they work.

Suggest using `as const` instead

See https://tduyng.medium.com/why-you-might-be-using-enums-in-typescript-wrong-6d9c5742db5a

```suggestion
// https://docs.sourcepoint.com/hc/en-us/articles/4405397484307-Event-callbacks-CMP#h_01FTY32EGZ7H3SS192MEE6SNCP
export const SourcePointChoices = {
  AcceptAll: 11,
  RejectAll: 13,
  Dismiss: 15
} as const;
```

libs/@guardian/libs/src/consent-management-platform/sourcepoint.ts
coldlink: Screenshot 2024-12-17 at 13 40 04

Yeah unfortunatley this doesn't work. As you cant use a type as a value.

pvighi: ah you are right,  I see now that I made a mistake in the way I defined the enum in  my scratchpad test.
I wonder what we get from enums if we are never using it as a type anywhere though.

Unless we are passing around variables of type `SourcePointChoiceTypes` wouldn't it make more sense to have regular number constants ?
Maybe applying some scala thinking to a typescript project but I would only bother with enums if I'm planning to convert the numbers we get from SourcePoint into values of the enum and pass that around.

libs/@guardian/libs/src/consent-management-platform/lib/sourcepointConfig.ts
akinsola-guardian: Thanks for this. Just updated the PR.

coldlink (approved):

PR #1847, opened 2024-12-10
Stop loading the cmp if user is on info page

akinsola-guardian:
## What are you changing?

- Preventing the CMP banner from loading on informational pages.
- Configuring the excludePage as a targeting parameter for Sourcepoint.

## Why?

- The CMP banner should not show on info pages.

### DIFF
--- a/.changeset/wise-dancers-travel.md
+++ b/.changeset/wise-dancers-travel.md
@@ -1,5 +1,5 @@
 ---
-'@guardian/libs': minor
+'@guardian/libs': major
 ---
 
-Stop showing CMP on info pages
+The CMP will no longer show for pages with `window.guardian.config.page.section: 'info' | 'help'`
--- a/.changeset/wise-dancers-travel.md
+++ b/.changeset/wise-dancers-travel.md
@@ -1,0 +1,5 @@
+---
+'@guardian/libs': minor
+---
+
+Stop showing CMP on info pages
--- a/libs/@guardian/libs/src/consent-management-platform/sourcepoint.ts
+++ b/libs/@guardian/libs/src/consent-management-platform/sourcepoint.ts
@@ -75,7 +75,8 @@ export const init = (framework: ConsentFramework, pubData = {}): void => {
 
 	log('cmp', `framework: ${framework}`);
 	log('cmp', `frameworkMessageType: ${frameworkMessageType}`);
-	const pageSection = window.guardian?.config?.page?.section as string;
+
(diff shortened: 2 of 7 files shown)


libs/@guardian/libs/src/consent-management-platform/sourcepoint.ts
```diff
@@ -195,11 +198,15 @@ export const init = (framework: ConsentFramework, pubData = {}): void => {
 
 	switch (framework) {
 		case 'tcfv2':
-			window._sp_.config.gdpr = {
-				targetingParams: {
-					framework,
-				},
-			};
+			{
+				const pageSection = window.guardian?.config?.page?.section as string;
```
AshCorr, line 202: Could `isExcludedFromCMP` be updated to accept `string | undefined` so we don't need this typecast? Feels like we shouldn't have to circumvent Typescript here!

AshCorr (approved):

.changeset/wise-dancers-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/libs': minor
+---
+
+Stop showing CMP on info pages
```
oliverabrahams, line 5:
```suggestion
The CMP will no longer show for pages with `window.guardian.config.page.section: 'info' | 'help'` 
```

.changeset/wise-dancers-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/libs': minor
```
oliverabrahams, line 2: I would make this a major change. As it is now not showing for info pages a consumer may have to change their tests etc.

oliverabrahams (commented):
Left a few comments to make the changes clearer for consumers

.changeset/wise-dancers-travel.md
akinsola-guardian: Thanks

oliverabrahams (approved):
Looks good

sndrs:
sorry @akinsola-guardian, i'm late to this (i was off), but one issue with this approach is that, afaik, there's no guarantee that that info will always be available on that attribute of a global variable e.g. https://github.com/guardian/csnx/discussions/444.

that's def an internal implementation detail of DCR (and other teams coming to rely on it are a really good reason to remove it 😆).

if we need to not call the CMP on certain pages, it would be more robust to have those pages not initialise it themselves, rather than the CMP 'know' whether it should initialise

PR #1849, opened 2024-12-12
Make keyboard appear on mobile and Ipad

oliverabrahams:
## What are you changing?

add a wrapper to the grid which adds `contenteditable` property so that the keyboard appears on mobile - we still need to prevent default on input and select to stop you from actually editing the grid components - We want to edit the content using the keydown not the input to change the value of the cells. We need to prevent default on the select start because we do not want the text to be selected as we do not want the text selected either.

## Why?

- Make the keyboard appear on mobile and Ipad


sndrs (approved):
this is really excellent ollie, great work

PR #1853, opened 2024-12-13
use input to capture key press on android

oliverabrahams:
## What are you changing?

-  Adding an input which is not visible but is over the entire crossword to capture the input from the letters added to the grid

## Why?

- I chose this method as opposed to the floating input as when doing a clue the keyboard and screen jump around following the input this felt quite clumsy on mobile

### DIFF
--- a/libs/@guardian/react-crossword/src/components/Grid.tsx
+++ b/libs/@guardian/react-crossword/src/components/Grid.tsx
@@ -180,8 +180,10 @@ export const Grid = () => {
 				: keyDownRegex.test(key) && key.toUpperCase();
 
 			if (value) {
+				// This mimics moving to a new input cell after typing a letter
 				inputRef.current?.blur();
 				inputRef.current?.focus();
+
 				updateCell({
 					x: currentCell.x,
 					y: currentCell.y,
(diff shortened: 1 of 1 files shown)


libs/@guardian/react-crossword/src/components/Grid.tsx
```diff
@@ -387,16 +373,16 @@ export const Grid = () => {
...
-			suppressContentEditableWarning={true}
 			ref={gridWrapperRef}
 			css={css`
+				position: relative;
 				cursor: pointer;
-				caret-color: transparent;
 				width: 100%;
 				max-width: ${width}px;
 				max-height: ${height}px;
+				-webkit-tap-highlight-color: transparent;
```
sndrs, line 383: comment to describe why this is here?

libs/@guardian/react-crossword/src/components/Grid.tsx
```diff
@@ -166,12 +168,44 @@ export const Grid = () => {
...
+				return;
+			}
+			const direction = currentEntryId?.includes('across') ? 'across' : 'down';
+			const key = event.target.value.toUpperCase();
+			const value = cheatMode
+				? cells.getByCoords({ x: currentCell.x, y: currentCell.y })?.solution
+				: keyDownRegex.test(key) && key.toUpperCase();
+
+			if (value) {
+				// This mimics moving to a new input cell after typing a letter
```
sndrs, line 183: can this describe why we need to do this?

sndrs (commented):
couple of minor comments but amazing digging!

sndrs (approved):
couple of minor comments but amazing digging!

PR #1866, opened 2024-12-19
A11y crossword grid

oliverabrahams:
## What are you changing?
- Adds formatCluesForScreenReader as a utility function.
- Add aria-label for input to improve accessibility.
- Add a `grid` and `row` role parent components and make the child a `gridcell` in line with docs https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/gridcell_role


sndrs (approved):
love it!!

PR #1885, opened 2025-01-07
Add sticky clue

oliverabrahams:
## What are you changing?

- Add a sticky component to the crossword. It can go on top of the grid and be used to see the current clue when on mobile

## Why?

- Users do not want to scroll down on mobile to be reminded of the clue. This means the user can see the clue they are currently on as well as any other clues


libs/@guardian/react-crossword/src/components/StickyClue.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+		),
+	],
+};
+
+export default meta;
+
+type Story = StoryObj<typeof StickyClue>;
+
+export const Across: Story = {
+	args: {},
```
jamesmockett, line 30: You shouldn't need to specify `args` as it's empty anyway.

oliverabrahams: Good point

libs/@guardian/react-crossword/src/components/StickyClue.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+	const stickyClue = css`
+		top: 0;
+		position: sticky;
+		display: flex;
+		z-index: 1;
+		min-height: 3.5em;
+		align-items: start;
+		${textSans12};
+		background: white;
```
jamesmockett, line 28: Can this use one of the existing colours from the theme? (Or perhaps it needs its own entry?)

libs/@guardian/react-crossword/src/components/StickyClue.tsx
```diff
@@ -0,0 +1,59 @@
...
+							font-weight: bold;
+							padding-right: 0.625em;
+							text-transform: capitalize;
+						`}
+					>
+						{entry.id.split('-').join(' ')}
+					</span>
+					<span
+						aria-hidden="true"
+						css={css``}
```
jamesmockett, line 48: This can be removed as it's empty.

libs/@guardian/react-crossword/src/components/StickyClue.tsx
oliverabrahams: Ohhh yess

jamesmockett (approved):

PR #1887, opened 2025-01-09
Update readme.md

shtukas:
## What are you changing?
Updating the readme because what to do wasn't obvious to me. Now it's more clear.

### DIFF
--- a/libs/@guardian/prettier/readme.md
+++ b/libs/@guardian/prettier/readme.md
@@ -16,7 +16,12 @@ npm install --save-dev @guardian/prettier
 
 ## Usage
 
-prettier has two configuration files: .prettierignore, which is layed out like git's .gitignore and .prettierrc.json which should have the following contents to call the right instructions.
+Prettier can take two configuration files: 
+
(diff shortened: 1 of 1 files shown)


libs/@guardian/prettier/readme.md
```diff
@@ -16,6 +16,8 @@ npm install --save-dev @guardian/prettier
 
 ## Usage
 
+prettier has two configuration files: .prettierignore, which is layed out like git's .gitignore and .prettierrc.json which should have the following contents to call the right instructions.
```
sndrs, line 19: thanks for this pascal. i added a few links to hopefully help this is still not clear:

```suggestion
Prettier can take two configuration files: 

- `.prettierignore` which works like git's `.gitignore`
- `.prettierrc.yml` (or [an equivalent](https://prettier.io/docs/en/configuration.html)) which provides settings to prettier

To use this package, your prettier config should [reference it directly](https://prettier.io/docs/en/sharing-configurations#using-a-shareable-config):
```

sndrs (approved):

PR #1891, opened 2025-01-13
Crossword inputs in all cells Grid A11y

oliverabrahams:
## What are you changing?

- The way the Grid.tsx component works
previously, there was an svg which was a visual representation and then a single input that stored the information about the cell.
Now, Each Cell component is an Aria role [cell](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/cell_role). This means that even with the browser/ screen reader navigation, the user can move the focus around the different cells.
- Make all the crossword cells inputs. This improves navigation of the screen reader as all the cells can be focused and have content all the time, not just when they are the current cell.
- Handle focusing the correct cell using the ID rather than refs, now all of the cells have inputs we just need to find the correct one and focus it rather than updating the ref and then finding that ref.

## Why?

- With the single input method the browser / screen reader navigation would move focus from the cell out into the table.
- With this new method the platform already has all the cells in the DOM therefore it can navigate using the platform navigation.


arelra (approved):
Nice improvement ✨ 

It might be helpful in the PR description to have a before and after snapshot of the DOM to help visualise the changes to the underlying crossword cells.

PR #1893, opened 2025-01-15
Grid a11y handle click

oliverabrahams:
## What are you changing?

- When clicking a cell it changes the current entry to the other direction if it exists.

## Why?

- This is the expected behaviour


libs/@guardian/react-crossword/src/components/Grid.tsx
```diff
@@ -134,6 +134,25 @@ export const Grid = () => {
...
+				cell.y === currentCell.y &&
+				cell.group?.length
+			) {
+				const otherEntry = cell.group.find(
+					(entryId) => entryId !== currentEntryId,
+				);
+				if (otherEntry) {
+					setCurrentEntryId(otherEntry);
+					return;
+				}
```
arelra, line 150: Just on the naming. We are finding `otherEntry` but then using it to `setCurrentEntryId` ?

```suggestion
				const otherEntry = cell.group.find(
					(entryId) => entryId !== currentEntryId,
				);
				if (otherEntry) {
					setCurrentEntry(otherEntry);
					return;
				}
```

arelra (approved):

libs/@guardian/react-crossword/src/components/Grid.tsx
oliverabrahams: Good point. I renamed to `otherEntryId`

PR #1894, opened 2025-01-15
Fix for stored state. 

oliverabrahams:
If the state received is invalid then set the state to be the default state and return the default state.
- Currently it is returning the default state provided as the state but not resetting the state to the default value.

### DIFF
--- a/libs/@guardian/react-crossword/src/hooks/useStoredState.ts
+++ b/libs/@guardian/react-crossword/src/hooks/useStoredState.ts
@@ -2,7 +2,6 @@
  * Wrapper for https://github.com/astoilkov/use-local-storage-state that:
  *
  * - adds a `validator` option
- * - makes `defaultValue` required
  * - provides our own serializer to keep the data stored in localStorage in a
  *   format that matches the format used by @guardian/libs#storage
  *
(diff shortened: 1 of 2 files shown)


libs/@guardian/react-crossword/src/hooks/useStoredState.ts
```diff
@@ -53,16 +53,17 @@ export function useStoredState<
...
 		// If the state is valid, return it (now properly typed).
 		if (validator(state)) {
 			return state;
 		}
 
-		// The state is invalid, so return the default value (which may be
-		// undefined).
+		// The state is invalid, set the state to the default value and return the default value
+		// (which may be undefined).
+		setState(defaultValue);
```
arelra, line 64: Minor suggestion - I wonder if we are mixing concerns here i.e. validating state and setting state. Could this function return `T | undefined` where `undefined` denotes invalid state. The it's up to the calling function to set the state to `defaultValue` if `undefined` ?

arelra (approved):

libs/@guardian/react-crossword/src/hooks/useStoredState.ts
oliverabrahams: Ahh yes that does seem better. I think it makes more sense to return an undefined state and just handle the validation in the function.

PR #1895, opened 2025-01-15
Grid a11y format label

oliverabrahams:
## What are you changing?

- Adding an aria label that describes the cell you are on.
    - The label is in the format  `Letter 2 of 4-across: Life is in a mess (5 letters). Also, letter 1 of 5-down Life is always in a mess (2 letters).`

## Why?

- This will give screen readers all the info needed to fill in a cell.


libs/@guardian/react-crossword/src/components/Grid.tsx
```diff
@@ -17,15 +19,50 @@ import { useProgress } from '../context/Progress';
...
+}: {
+	entry: CAPIEntry;
+	cell: CellType;
+	additionalEntry?: boolean;
+}): string => {
+	if (entry.direction === 'across') {
+		return `${additionalEntry ? 'Also, letter' : 'Letter'} ${cell.x + 1 - entry.position.x} of ${entry.length}. ${entry.id}. ${formatClueForScreenReader(entry.clue)}`;
+	} else {
+		return `${additionalEntry ? 'Also, letter' : 'Letter'} ${cell.y + 1 - entry.position.y} of ${entry.length}. ${entry.id}. ${formatClueForScreenReader(entry.clue)}`;
+	}
```
arelra, line 65: Is this equivalent?

```suggestion
const cellPosition = entry.direction === 'across'
    ? String(cell.x + 1 - entry.position.x)
    : String(cell.y + 1 - entry.position.y);
return `${additionalEntry ? 'Also, letter' : 'Letter'} ${cellPosition} of ${entry.length}. ${entry.id}. ${formatClueForScreenReader(entry.clue)}`;
	
  ```

Comment (oliverabrahams) [reply libs/@guardian/react-crossword/src/components/Grid.tsx:65]: Oh nice yea that is 👍

File: libs/@guardian/react-crossword/src/utils/stripHTML.ts [hunk 2]
```diff
@@ -0,0 +1,4 @@
+export const strip = (html: string): string => {
+	const doc = new DOMParser().parseFromString(html, 'text/html');
```
arelra, line 2: I have a concern that using `DomParser` on (I assume) every clue will be costly performance wise.

https://webperf.tips/tip/uber-eats-dom-parsing/

Is there another way we can grab the text we require from the data model that doesn't require parsing the DOM?

oliverabrahams: Interesting, The clues come with some formatting in HTML but i wanted to remove it for the screen reader as it says the tags out loud. I could use a regex as i guess the html will be quite predictable, or do this once when the crossword loads in the parseCrosswordData function. https://github.com/guardian/csnx/blob/main/libs/%40guardian/react-crossword/src/utils/parseCrosswordData.ts . Which do you think?

oliverabrahams: I am going to add the description to the data model for the cell.

arelra: Sorry I missed this. That sounds ideal - thanks!

jamesmockett, line 2: I wonder if it would be simpler to use a regex to do this as we're only really concerned about basic formatting tags like `<i>` or `<b>`? There's an example in the DCR codebase:

https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/components/TextBlockComponent.tsx#L45-L51

Now this is only happening once rather than every time the component re-renders I don't think it matters too much though.

oliverabrahams: I am not sure which is better. The tags should be simple but we do not know what the clue will be so if there is a clue like '2 < 3 and 5 > 1' the regex would cut out the '< 3 and 5 >'

sndrs: this seems like a good approach to me. you get all the work the browser vendor has put into parsing the HTML string for free, including all the potential weird edge cases etc. it's [notoriously](https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags) hard to get right with regex. 

my guess is that this is a pretty performant API (given is the at the heart of what a browser does) but it might be worth seeing if memoising responses helps if `strip` is called a lot.

(i'd also call it `stripHTMLTags`, it's not clear what's being stripped by itself)

oliverabrahams: The issue with performance is from what Ravi said [here](https://github.com/guardian/csnx/pull/1895#discussion_r1924037682). By stripping the tags initially and adding the description to the data object in the initial parsing of the crossword, It is only running 1 * number of cells and not on re-renders. I think the performance should not be affected now and this is more reliable at striping what we don't want

PR #1902, opened 2025-01-23
Handle Crossword input for IMEs / Android native keyboard.

oliverabrahams:
## What are you changing?

- Adding an onInput event handler to the crossword cells. To remove or add letters if the onKeydown event doesn't have the required information.

## Why?

- IMEs which includes the android keyboard dont add information about which key was pressed.
    - To handle this in react we need to use the native event. 
    - The native event has the information we need to know if we are supposed to add or remove a letter
Resources which explain the problem
   - https://github.com/facebook/react/issues/14512
   - https://clark.engineering/input-on-android-229-unidentified-1d92105b9a04


jamesmockett (approved):
Looks good 👏

PR #1910, opened 2025-01-27
Back to regex

oliverabrahams:
## What are you changing?

- Using regex to strip tags rather than DOMParser

## Why?

- This code needs to run on the server where there is no DOMParser and jsdom is 3.1mb.
- We know which tags can be possible from the docs here https://crossword.info/docs/schema-doc.html#h464208646 therefore, we can be confident that the regex will be sufficient.


jamesmockett (approved):
This is great 👏 Definitely simplifies things now we know there's only a limited set of allowed tags.

PR #1923, opened 2025-02-03
Use specific corepack version

andrewHEguardian:
## What are you changing?

- Use a specific version of corepack

## Why?

- The build is failing, see https://github.com/pnpm/pnpm/issues/9014. This is a proposed fix also used in https://github.com/guardian/support-service-lambdas/pull/2666


oliverabrahams (approved):
Looks great 👍

PR #1925, opened 2025-02-04
Allow user to navigate clues without being taken to grid using arrow keys

oliverabrahams:
## What are you changing?

- Previously when moving between the clues using the arrow keys you would also select them. Now you can navigate focus with the arrow keys and then use the spacebar or the enter key to select the clue and focus the grid.

## Why?

- This is a better experience for keyboard navigation


jamesmockett (approved):
This is a nice little improvement ✨

PR #1926, opened 2025-02-04
Crossword print styles

jamesmockett:
## What are you changing?

Adds basic print styles to crossword component

## Why?

So crossword is visible and legible when printed, doesn't show unnecessary elements and makes best use of available space

## Screenshots

| Before      | After      |
| ----------- | ---------- |
| ![before][] | ![after][] |

[before]: https://github.com/user-attachments/assets/860409c9-f234-4288-8a8e-14cca33f3111
[after]: https://github.com/user-attachments/assets/0ac52d81-3af0-4e19-975d-19f8b71c1a69


oliverabrahams (approved):
Looks great

PR #1927, opened 2025-02-05
Support Ophan data- attributes in Source

SiAdcock:
## What are you changing?

- Add support for the data-link-name  attribute in Source Buttons and Link components
- Add support for the data-component attribute in all Source components 

## Why?

We use Ophan to track clicks on and renders of many of our components in our applications. This change adds first class support of Ophan tracking to Source components. Passing values for these new props to the components allows Ophan's Tracker JS to identify and track these components. These props are optional – nothing happens if the props are omitted.

### DIFF
--- a/.changeset/lucky-nails-swim.md
+++ b/.changeset/lucky-nails-swim.md
@@ -1,0 +1,5 @@
+---
+"@guardian/source": minor
+---
+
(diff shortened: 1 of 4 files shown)


SiAdcock:
Ignore me, this is already possible because these components accept all valid html attributes, including data-* attributes

PR #1929, opened 2025-02-06
Fix for crossword android keyboard

oliverabrahams:
## What are you changing?

- Making the input type: `insertCompositionText` also input on the crossword - https://w3c.github.io/input-events/#input-event-order-during-composition

## Why?

currently not working on Android


jamesmockett (approved):
⌨️

PR #1935, opened 2025-02-10
Add Ophan tracking to crossword

jamesmockett:
## What are you changing?

Add `data-link-name` attributes to crossword controls so that interaction is captured in Ophan

### DIFF
--- a/libs/@guardian/react-crossword/src/components/AnagramHelper.tsx
+++ b/libs/@guardian/react-crossword/src/components/AnagramHelper.tsx
@@ -83,6 +83,7 @@ export const AnagramHelper = () => {
 				min-height: fit-content;
 				z-index: 2;
 			`}
+			data-link-name="Anagram Helper"
 		>
 			<div
 				css={css`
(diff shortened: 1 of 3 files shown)


SiAdcock (approved):
This LGTM thanks @jamesmockett 👍 Have you tested in DCR to make sure the Ophan events are firing correctly, or are we testing that in PROD?

jamesmockett:
> Have you tested in DCR to make sure the Ophan events are firing correctly, or are we testing that in PROD?

I'll make a canary release from this branch and test in DCR before merging.

There's one button that doesn't currently have any tracking as the anagram helper UI is subtly different. In the new version you have to click _Start_ after entering the letters:

Screenshot 2025-02-11 at 11 54 57

In the existing version you click _Shuffle_:

Screenshot 2025-02-11 at 11 54 27

It might make sense for the _Start_ button to have the same tracking ID as the _Shuffle_ button as it effectively performs the initial shuffle?

SiAdcock:
> It might make sense for the Start button to have the same tracking ID as the Shuffle button as it effectively performs the initial shuffle?

Yeah I think that's reasonable. I don't expect much risk of data loss, nobody is actively keeping track of the usefulness of the anagram helper.

oliverabrahams (approved):

PR #1936, opened 2025-02-10
Revert "Use specific corepack version"

andrewHEguardian:
Testing if we still need this specific version specified


andrewHEguardian:
image
Yes, still required

PR #1938, opened 2025-02-11
Fixing issue with re rendering using local storage

oliverabrahams:
## What are you changing?

- Adding a new progress state which uses react state to trigger rerenders

## Why?

- 	The localStorage state (managed by useLocalStorageState) does not use React state
	so updates to it do not always trigger a re-render - this behavior is particularly noticeable in Preact.
	To ensure the UI is kept up-to-date, we maintain a separate React state (`progress`) that mirrors
	the localStorage state and forces the necessary re-renders.


jamesmockett (approved):
Nice work 👏

PR #1939, opened 2025-02-12
Add Prebid adaptors to vendor ID list

cemms1:
## What are you changing?

- Adds individual Prebid adaptors to the vendor list so that we can use them in the `getConsentFor` function

## Why?

- We want to start checking consent for each Prebid adaptor individually rather than letting them handle the consent themselves. Adding each Prebid adaptor to the vendor list allows us to only allow interactions with the bidders when consent is explicitly granted for a vendor rather than relying on a global Prebid custom vendor and hoping each adaptor manages consent appropriately.

### DIFF
--- a/.changeset/olive-bobcats-grab.md
+++ b/.changeset/olive-bobcats-grab.md
@@ -2,4 +2,4 @@
 '@guardian/libs': minor
 ---
 
-Add Prebid adaptors to TCF vendor ID list
+Add Prebid adaptors to CMP TCF vendor ID list


theemilyturner:
Could you please update ix to index exchange, and also confirm the full name of ttd as I can't see what vendor that one is?

theemilyturner:
According to Charlie Vicente, Triple Lift, Inc should only be active in AU and US.

Amouzle:
> According to Charlie Vicente, Triple Lift, Inc should only be active in AU and US.

Oh yes! That was my bad. Thanks!!

Jakeii (approved):
If Alex is happy I'm happy!

andrewHEguardian (approved):

.changeset/olive-bobcats-grab.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/libs': minor
+---
+
+Add Prebid adaptors to TCF vendor ID list
```
andrewHEguardian, line 5: You could add that this is a CMP specific change to the changelog

andrewHEguardian: Yes, sorry for the confusion I did mean this changeset 👍

PR #1940, opened 2025-02-12
Update `@guardian/libs` instructions for linking

cemms1:
## What are you changing?

- Updates the paths in the commands for linking and unlinking in the `@guardian/libs` README file

## Why?

- I found the original commands didn't work for me - though that could just have been due to differences between `pnpm link` and `npm link`?


oliverabrahams (approved):
Nice spot!

PR #1941, opened 2025-02-12
Add check to cell and a display to show if the cell is correct 

oliverabrahams:
## What are you changing?

- Adding a slash through checked letters that are incorrect.
- Adding a label to incorrect letters to make it clear to screen reader users which letters are incorrect.

Screenshot 2025-02-12 at 11 34 19

## Why?

- Screen reader users need to also know which letters are incorrect. 
- It is not accessible to have letters just be removed from the grid - it makes some sense visually but not to screen reader users.


jamesmockett (approved):
Nice. This makes sense visually and for screenreader users 👏

PR #1943, opened 2025-02-12
Crossword controls A11y

oliverabrahams:
## What are you changing?

- Update the clue controls to have a label which tells which word is going to be checked / revealed / cleared
- Labels the Anagram Helper close button
- Auto focuses text input field when Anagram Helper opened

## Why?
- Improve a11y of the buttons

### DIFF
--- a/libs/@guardian/react-crossword/src/components/Controls.tsx
+++ b/libs/@guardian/react-crossword/src/components/Controls.tsx
@@ -66,7 +66,6 @@ const ClearClue = (props: ButtonProps) => {
 			aria-label={`Clear ${currentEntryId ? currentEntryId.split('-').join(' ') : 'word'}`}
 			data-link-name="Clear this"
 			aria-live={'off'}
-			role={'menuitem'}
 			{...props}
 		>
 			Clear Word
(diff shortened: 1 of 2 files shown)


libs/@guardian/react-crossword/src/components/Controls.tsx
```diff
@@ -149,6 +167,8 @@ const AnagramHelper = (props: ButtonProps) => {
 			onClick={toggleAnagramHelper}
 			data-link-name="Show anagram helper"
 			{...props}
+			role="none"
+			aria-hidden="true"
```
jamesmockett, line 171: There's no need for both as `aria-hidden` will remove the element from the accessibility tree so it's role won't be exposed anyway: https://www.scottohara.me/blog/2018/05/05/hidden-vs-none.html

libs/@guardian/react-crossword/src/components/Controls.tsx
```diff
@@ -424,6 +443,11 @@ export const Controls = () => {
 					}
 					return null;
 				})}
+				<AnagramHelper
+					key={'anagram-helper'}
+					disabled={disableClueControls}
+					tabIndex={-1}
```
jamesmockett, line 449: Is this going to prevent all users from tabbing to the _Anagram Helper_ button? (I night be misunderstanding as some attributes are being set here, and others directly on `ClueButton` further up.)

jamesmockett (approved):

PR #1945, opened 2025-02-14
Performance sweep

oliverabrahams:
## What are you changing?

- Improving performance of crosswords by adding in memoisation of components and only rerendering the cells and clues that have changed rather than all of them.

## Why?

- It is important that the crossword works on devices with slower computation

Co-authored with @jamesmockett 

Improves the rendering time from about 30ms to 4-7ms


andrewHEguardian (approved):
I don't have the full context having not seen this code before, but the memoisation looks great! How did you measure the performance?

oliverabrahams:
> I don't have the full context having not seen this code before, but the memoisation looks great! How did you measure the performance?

using React dev tools, the profiler mostly

PR #1957, opened 2025-02-17
Re release packages as latest

oliverabrahams:
## Why?

- Accidentally merged a canary PR with main before the packages could be released for real 
 There is a race condition where if a branch with the canary tag merges with main before the release then the new versions can be released as canaries. https://github.com/guardian/csnx/pull/1823#issuecomment-2663308959


andrewHEguardian (approved):
🐥

PR #1958, opened 2025-02-17
Add noop to crosswords

oliverabrahams:
React is erroring when there is an input with no onChange or onInput function


andrewHEguardian (approved):
👍

PR #1966, opened 2025-02-20
react-crossword A11y improvements

oliverabrahams:
## What are you changing?

- Reduce the verbosity of the aria labels on the grid cells
- Remove the application role from the parent div as it is making the crossword not very visible to screen readers

## Why?

- We have received feedback from a user that the labels are too verbose and make the crossword very difficult to use. The  information that was previously being read out can in fact be inferred by the position on the grid and the previous navigation context.

**previously** 
Every cell had information about what clue it was part of and how many letters there were in the word
Every Cell -> `Letter 2 of 4-across: Life is in a mess (5 letters). Also, letter 1 of 5-down Life is always in a mess (2 letters).`

**now** 
Only the cells at the start of a word have info about the clue and length. This gives the context for the surrounding letters. This should reduce the overload of information read out.
First Cell -> `4-across: Life is in a mess (5 letters).`

### DIFF
--- a/libs/@guardian/react-crossword/src/components/Cell.tsx
+++ b/libs/@guardian/react-crossword/src/components/Cell.tsx
@@ -1,4 +1,5 @@
 import { css } from '@emotion/react';
+import { isUndefined } from '@guardian/libs';
 import { textSans12 } from '@guardian/source/foundations';
 import type { FormEvent, KeyboardEvent, SVGProps } from 'react';
 import { useEffect, useRef } from 'react';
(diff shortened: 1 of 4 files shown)


jamesmockett (approved):

bryophyta:
Seems like a good idea to me given the feedback we've received so far 👍 

There's the possibility of putting some of the info that's removed here into an `aria-details` value. In theory it can provide supplementary information, without getting included in the accessible description of an element. I don't know how widely `aria-details` is used though, so we'd probably want to do a bit more research and testing before using it.

PR #1976, opened 2025-02-25
Send propertyHref to Sourcepoint and reinstate CORP_FLAG.

akinsola-guardian:
## What are you changing?

- Sending propertyHref to Sourcepoint as well as propertyId
- Reinstating the CORP_FLAG query param to allow for testing.

## Why?

- On recommendation from Sourcepoint, as their legacy system added a propertyHref which caused issue with loading the correct banner and vendor list.
- Query param was previously removed but more testing is required.


pvighi (approved):
not super familiar with this but it looks good to me

AshCorr (approved):

PR #1978, opened 2025-02-27
Refactor: "Check Word" button clears entire word

DanielCliftonGuardian:
## What are you changing?

This PR updates the "Check Word" button behaviour to match the functionality of the previous crossword. The button now clears all entered letters in the current word, regardless of correctness. This aligns with user expectations from the previous version.

### DIFF
--- a/libs/@guardian/react-crossword/CHANGELOG.md
+++ b/libs/@guardian/react-crossword/CHANGELOG.md
@@ -1,5 +1,16 @@
 # @guardian/react-crossword
 
+## 4.0.0
+
+### Major Changes
+
+- aeb7ed2: Extends theme to support additional elements and renames some existing keys for consistency and clarity:
+
+  - Adds `gridTextColor` to theme to allow text in the grid to be themed independently of other text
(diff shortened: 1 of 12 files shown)


SiAdcock:
There's three reasons I think we should clear incorrect letters instead of striking them out:

1) The contrast of the strikethrough is hard to get right. Too much contrast and it obscures the letter. Too little and it is too feint for some users to discern. [Some](https://www.theguardian.com/crosswords/quick/17099#comment-170771341) [users](https://www.theguardian.com/crosswords/quick/17102#comment-170816113) are already struggling with this.
2) if you have a lot of crossed out letters and only one or two that are correct, it's really hard to distinguish. There's a lot of visual noise. (see screenshot)
3) leaving the letters in place means users have a harder time "imagining" a different word in that space. They have to manually delete the letters, which is a bit of a drudge when you're just trying to relax / have fun.

Screenshot 2025-02-27 at 16 09 09

jamesmockett:
This will need a changelog adding before merging now that we've published the crossword package and are no longer relying on canary releases.

SiAdcock (approved):

PR #1980, opened 2025-02-27
Support scrolling clue into view when selected

jamesmockett:
## What are you changing?

Adds `scrollToSelected` prop to `Clue` component to automatically scroll clue into view when selected

## Why?

On DCR we want to allow the clue list to be scrolled independently of the rest of the page on tablet sized viewports so readers can see all of the clues whilst keeping the crossword grid in view. As part of this, the clue list should automatically scroll to the relevant clue when a cell is selected on the grid.


SiAdcock (approved):
👍

PR #1993, opened 2025-03-05
Add print background colour and cell border for Firefox

SiAdcock:
## What are you changing?

- Add a print-only background colour to each cell
- Add a print-only stroke colour to each cell 

## Why?

By default, Firefox doesn't support printing background colours. This behaviour must be [enabled using a setting in the print dialog](https://superuser.com/questions/1113516/is-it-possible-to-view-a-page-in-firefox-using-print-media-type). 

We can get around it using the `fill` property, which is currently set to `transparent` for the black cells. We can use the `stroke` property to set the border colour between cells.

## Screenshots

**Before**

Screenshot 2025-03-05 at 17 57 25

**After**
Screenshot 2025-03-05 at 21 25 29


jamesmockett (approved):
✨

PR #2005, opened 2025-03-12
Extend crossword theming

jamesmockett:
## What are you changing?

- Adds `gridTextColor` to theme to allow text in the grid to be themed independently of other text
- Explicitly applies `textColor` to other elements rather than rely on inheritance or `currentColor`
- Renames `selectedColor` and `connectedColor` to make it clear they apply to background colours
- Renames `focusedClueBackgroundColour` to standardise spelling as `color`

## Why?

Primarily to support dark mode where some elements need to change from dark to light independently of the crossword grid.

## Screenshots

Screenshot 2025-03-12 at 17 46 56


oliverabrahams (approved):
Looks good to me!

PR #2008, opened 2025-03-14
Add monospace font Courier New to storybook

oliverabrahams:
## What are you changing?

- Adding a specific monospace font to storybook

## Why?

- To stop consistent changes to the font which have to be accepted over and over.


andrewHEguardian (approved):
🤞

PR #2020, opened 2025-03-19
Remove usage of Array.at() - it is not supported in older browser versions

oliverabrahams:
remove usage of Array.at() - it is not supported in older browser versions


jamesmockett (approved):
✨

PR #2026, opened 2025-03-21
Crosswords - change working direction when moving.

oliverabrahams:
## What are you changing?

- Update the working direction when you move around the grid with the arrows.
- Fix for chromatic where the id of the local storage has changed to `crosswords.${id}`
## Why?

- For crosswords where across / down clues are next to each other you want to change working direction when you navigate around the grid

## Before 

https://github.com/user-attachments/assets/b3a3adac-6cd9-4b65-bd32-432390e5542c

## After

https://github.com/user-attachments/assets/d302d71a-1d0a-44de-9c41-73d0d1aca9a5


andrewHEguardian (approved):
Nice ⬆️

libs/@guardian/react-crossword/src/components/Grid.tsx
```diff
@@ -305,21 +305,30 @@ export const Grid = () => {
...
+			workingDirectionRef.current = direction;
+		}
+	};
 	const navigateGrid = useCallback(
 		(event: KeyboardEvent): void => {
 			let preventDefault = true;
 
 			switch (event.key) {
 				case 'ArrowUp':
+					updateWorkingDirection({ direction: 'down' });
```
andrewHEguardian, line 319: Would it be better if you had to press twice to change direction and move? ie first press just changes direction but not position. eg that way I could go to 6 down but still be on the first letter 
image

oliverabrahams: I did think this would be good also but for Aria grids the arrow keys should move you from one square to the next. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/grid_role This is why we implemented it like that maybe we could use another key to switch working direction... like `space` or something

PR #2039, opened 2025-03-25
Move through clues

oliverabrahams:
## What are you changing?

- Allows movement between clues using the `,` and `.` keys - Just a proof of concept. Not decided which keys are best for keyboard users.

https://github.com/user-attachments/assets/07608176-d32f-4639-bb11-e7799473b7c2


SiAdcock (approved):
Awesome! 👍

PR #2040, opened 2025-03-26
Bundle `use-local-storage-state` package with `@guardian/react-crossword`

jamesmockett:
## What are you changing?

- Extends base Rollup config to allow passing in options for `nodeExternals` plugin and CJS output
- Bundles `use-local-storage-state` package to allow it to be transpiled by consumers of crossword package

## Why?

The `use-local-storage-state` uses modern JS syntax, preventing crosswords from functioning in older browsers such as Safari 12

### DIFF
--- a/.changeset/warm-bears-sink.md
+++ b/.changeset/warm-bears-sink.md
@@ -1,0 +1,5 @@
+---
+'@guardian/react-crossword': minor
+---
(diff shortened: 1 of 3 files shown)


oliverabrahams (approved):
Fantastic

PR #2042, opened 2025-03-27
Add sunday-quick support to crossword player

andrew-nowak:
## What are you changing?

Adds sunday-quick support to the new crossword player

## Why?

So that it is available on the website


.changeset/slimy-shoes-listen.md
```diff
@@ -0,0 +1,6 @@
+---
+'@guardian/source-development-kitchen': minor
```
oliverabrahams, line 2: We don't need to change this crossword as it is not used anywhere. I added an issue to remove this https://github.com/guardian/dotcom-rendering/issues/13698

jamesmockett (approved):

PR #2059, opened 2025-04-17
Add `onbeforeload` listener to core-web-vitals

arelra:
## What are you changing?

-

## Why?

-


arelra:
Not required. The issue was in our data pipeline

PR #2060, opened 2025-04-17
Update Wireit to latest version

jamesmockett:
## What are you changing?

Updates Wireit to `0.14.12`

## Why?

Wireit was using a legacy GitHub caching API which was shutdown on 15 April, causing our builds to fail. The latest version has been updated to use the v2 caching API.

https://github.com/google/wireit/issues/1297


arelra (approved):
✨

PR #2065, opened 2025-04-29
Make goal copy in ticker configurable

LAKSHMIRPILLAI:
## What are you changing?

The PR is to make  the goal copy in ticker  configurable ie the word "goal" after 50,000 on this ticker

image

## Why?

 This is to change word "goal" after 50,000 on this ticker to "readers" to show that it's for supporter count

### DIFF
--- a/.changeset/sweet-wings-smell.md
+++ b/.changeset/sweet-wings-smell.md
@@ -1,5 +1,0 @@
----
-'@guardian/source-development-kitchen': major
----
-
-Make the goalCopy in TickerData configurable
--- a/.changeset/sweet-wings-smell.md
+++ b/.changeset/sweet-wings-smell.md
@@ -1,0 +1,5 @@
+---
+'@guardian/source-development-kitchen': major
+---
+
+Make the goalCopy in TickerData configurable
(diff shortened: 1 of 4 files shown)


.changeset/sweet-wings-smell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/source-development-kitchen': major
```
tomrf1, line 2: it's an optional field so this can be a minor release

libs/@guardian/source-development-kitchen/src/react-components/ticker/Ticker.tsx
```diff
@@ -18,6 +18,7 @@ interface TickerCopy {
 export interface TickerData {
 	total: number;
 	goal: number;
+	goalCopy?: string;
```
tomrf1, line 21: I think this belongs in the `TickerCopy` object

LAKSHMIRPILLAI: ok, I will change it to tickerCopy.

tomrf1 (approved):
great!

andrewHEguardian (approved):

PR #2067, opened 2025-04-30
Automatically calculate background hover color for Buttons

andrewHEguardian:
## What are you changing?

- Make background hover an optional property in Button themes, and calculate it based on the following ruleset: 

image
## Why?

- The Source design system currently applies a unique hover colour for each button colour. However, this approach isn't scalable when all button colours are customisable. This saves designers having to follow the ruleset to define a hover colour.


libs/@guardian/source/package.json
```diff
@@ -35,6 +35,7 @@
 		"verify-dist": "wireit"
 	},
 	"dependencies": {
+		"color-convert": "3.0.1",
```
SiAdcock, line 38: Just wanna check. It looks like `color-convert` _could_ add [5KB of JavaScript](https://bundlephobia.com/package/color-convert@3.0.1) to the bundled code. The consuming app might optimise this away (e.g. by tree-shaking), but it there's a chance it might not.

Can we try inlining the [relevant conversion functions](https://github.com/Qix-/color-convert/blob/master/conversions.js) to try and bring the size down?

Required functions:

- [`hex.rgb()`](https://github.com/Qix-/color-convert/blob/master/conversions.js#L685-L705)
- [`rgb.hsl()`](https://github.com/Qix-/color-convert/blob/master/conversions.js#L58-L112)
- [`hsl.rgb()`](https://github.com/Qix-/color-convert/blob/f53dad2f14e055e0bc7b4434ca13ec0435092172/conversions.js#L277-L318)
- [`rgb.hex()`](https://github.com/Qix-/color-convert/blob/f53dad2f14e055e0bc7b4434ca13ec0435092172/conversions.js#L674-L683)

andrewHEguardian: That seems like a sensible thing to do

libs/@guardian/source/src/react-components/utils/colourConversions.ts
```diff
@@ -0,0 +1,139 @@
+/*
+	Implementation from https://github.com/Qix-/color-convert
```
andrewHEguardian, line 2: @SiAdcock do I need to do anything special with the license?

SiAdcock: Oh good point! It's MIT, so pretty permissive as long as we [include the licence in the code](https://snyk.io/articles/what-is-mit-license/). Shall we copy-paste [the licence](https://github.com/Qix-/color-convert?tab=License-1-ov-file) into a comment in this file?

SiAdcock (approved):
This looks great, thanks @andrewHEguardian! 💎

PR #2068, opened 2025-04-30
Add new colour shades news.700 and sport.700

andrewHEguardian:
## What are you changing?

- Adding two new colour shades: news.700 and sport.700

## Why?

- The colours have been added to the Core Figma library


jamesmockett (approved):
Looks good! ✨

PR #2072, opened 2025-05-02
Add deprecation warning to Ophan types in libs

cemms1:
## What are you changing?

- Adds deprecation warning to Ophan types, signposting users towards `@guardian/ophan-tracker-js` instead

## Why?

- Continuing the work started in https://github.com/guardian/dotcom-rendering/pull/11438 to deprecate the Ophan types defined in `@guardian/libs` so that we can rely on one source of truth: `@guardian/ophan-tracker-js`

### DIFF
--- a/.changeset/chatty-mirrors-arrive.md
+++ b/.changeset/chatty-mirrors-arrive.md
@@ -1,0 +1,5 @@
+---
+'@guardian/libs': minor
+---
+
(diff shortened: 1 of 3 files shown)


jonflynng:
looks good

andrewHEguardian (approved):
👍

PR #2087, opened 2025-05-20
Add conditional for consent status

akinsola-guardian:
## What are you changing?

- Make consentStatus optional to handle undefined cases

## Why?

- This is causing issue as consentStatus is not always present on the gdpr object.

Tested in CODE.


codingWithRach (approved):
Looks sensible

PR #2088, opened 2025-05-20
Reattempt fixing react for shared eslint configs

andrew-nowak:
Brings back guardian/csnx#2079

#2079 caused some odd problems after merge - the CI run after merge succeeded, but then subsequent runs failed with cryptic error messages. Searching those messages on stack overflow brought me to https://stackoverflow.com/a/77464178 which suggests this is due to mismatching versions of a dependency, and to try running `pnpm dedupe`. I've done that, and it did manage to dedupe some things so that seems plausible. More concrete, once I'd deduped the dependency tree I started seeing some new linting failures coming from the crossword component, which needed fixing -- which seems to me to strongly suggest we've now got a working dependency tree! 🎉

### DIFF
--- a/.changeset/blue-points-occur.md
+++ b/.changeset/blue-points-occur.md
@@ -1,0 +1,5 @@
+---
+'@guardian/eslint-config': major
+---
+
+Bugfix to re-enable recommended react rules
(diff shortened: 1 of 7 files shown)


oliverabrahams (approved):
@andrewHEguardian @coldlink -- looks like this maybe an issue with the cache not invalidating when there are changes to the @guardian/eslint package

PR #2096, opened 2025-06-09
Migrate from GPP to Sourcepoint API

akinsola-guardian:
## What are you changing?

- Migrating from using the Global Privacy Platform (GPP) function to the Sourcepoint api to get consent in the US.

## Why?

- The GPP function is unreliable and can change it's api/implementation which leaves our solution exposed to potential failures in getting consent in the US.

Tested in CODE by:
-  checking the query parameters for the  google ad manager request are correctly set  - `rdp`( restrict downstream processes) and `pa` (personalised advertsing)
- running automated tests against CODE
- testing banner interactions

### DIFF
--- a/libs/@guardian/libs/src/consent-management-platform/types/usnat/index.ts
+++ b/libs/@guardian/libs/src/consent-management-platform/types/usnat/index.ts
@@ -34,7 +34,7 @@ interface UsnatCategories extends UsnatVendors {
 
 export interface UsnatData {
 	applies: boolean;
-	categories: UsnatCategories[];
-	vendors: UsnatVendors[];
+	categories?: UsnatCategories[];
+	vendors?: UsnatVendors[];
 	signalStatus: GPPSignalStatus;
 }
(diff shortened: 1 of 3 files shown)


libs/@guardian/libs/src/consent-management-platform/usnat/api.ts
```diff
@@ -1,19 +1,13 @@
...
-const api = (command: Command) =>
-	new Promise((resolve, reject) => {
-		if (window.__gpp) {
-			window.__gpp(command, (result, success) =>
-				success
-					? resolve(result)
-					: /* istanbul ignore next */
-						reject(new Error(`Unable to get ${command} data`)),
-			);
-		} else {
```
AshCorr, line 14: Should we keep this else? Seems like this promise might never resolve if `getUserConsents` doesn't exist on the window?

AshCorr (approved):

AshCorr (approved):

PR #2097, opened 2025-06-10
Include platform in core web vitals payload if set

tjmw:
## What are you changing?

- Update `InitCoreWebVitalsOptions` to include an optional `platform` string which will be sent to the events collection endpoint.

## Why?

- So that we can identify cwv metrics for different platforms.

### DIFF
--- a/libs/@guardian/core-web-vitals/README.md
+++ b/libs/@guardian/core-web-vitals/README.md
@@ -89,6 +89,21 @@ initCoreWebVitals(init);
 // should call log('dotcom', 'Core Web Vitals payload successfully queued […]')
 ```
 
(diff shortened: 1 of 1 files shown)


arelra (approved):
lgtm

Is it worth documenting if we don't pass platform that a default of 'dcr' will be applied by the events-collector?

andrewHEguardian (approved):

tjmw:
> > lgtm
> > Is it worth documenting if we don't pass platform that a default of 'dcr' will be applied by the events-collector?
> 
> Thanks @arelra, that's a great shout (and just documenting the new field in general)

This is done in 0504a24.

PR #2099, opened 2025-06-11
Remove deprecated getTCData command

codingWithRach:
## What are you changing?

- Remove `getTCData` from the list of valid commands for TCF v2
- Update the instructions for generating the fixture, to refer to `addEventListener` instead of `getTCData`
- Generate a new fixture file using the `addEventListener` command
- Update the snapshot by running `npx jest --updateSnapshot` from the `libs\@guardian\libs` folder so that tests pass

## Why?

- The `getTCData` command has been deprecated, and had already been replaced by `addEventListener` in the getTCData function


akinsola-guardian (approved):
LGTM. Thanks for clearing this up.

PR #2104, opened 2025-06-18
Remove calls to deprecated Ophan functions

codingWithRach:
## What are you changing?

- Replace use of OphanComponentEvent and OphanRecordFunction 

## Why?

- These are deprecated - the Ophan documentation says to use type definitions from `@guardian/ophan-tracker-js` instead

Tested locally

### DIFF
--- a/libs/@guardian/react-crossword/package.json
+++ b/libs/@guardian/react-crossword/package.json
@@ -44,7 +44,7 @@
 		"@storybook/test": "8.6.4",
 		"@types/jest": "29.5.8",
 		"@types/react": "18.2.79",
-		"dotenv": "16.5.0",
+		"dotenv": "17.0.0",
 		"eslint": "9.19.0",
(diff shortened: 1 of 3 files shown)


libs/@guardian/libs/package.json
```diff
@@ -27,6 +27,9 @@
 		"tsc": "wireit",
 		"verify-dist": "wireit"
 	},
+	"dependencies": {
+		"@guardian/ophan-tracker-js": "2.2.10"
+	},
```
coldlink, line 32: this should also be added to the `peerDependencies` too

rationale here: https://github.com/guardian/recommendations/blob/main/npm-packages.md#peerdependencies

codingWithRach: I'm not sure, but advice would be appreciated!

andrewHEguardian (approved):
Looks good!

PR #2134, opened 2025-07-31
Add Byline component to `stand`

andrewHEguardian:
## What are you changing?

- Adding the `Byline` component to the new stand lib. The only thing not moved over from `flexible-frontend` is the tests. We will rewrite these in playwright to avoid difficulties with react-testing-library and jsdom

## Why?

- This is where it will call home and be able to be used across tools


andrewHEguardian:
We don't have a Chromatic project for `stand` yet, we should do that in a future PR

coldlink (approved):

PR #2143, opened 2025-08-05
chore(stand): prepare byline component for export

coldlink:
## What are you changing?

-

## Why?

-


libs/@guardian/stand/package.json
```diff
@@ -54,16 +45,47 @@
 		"storybook": "8.6.4",
 		"ts-jest": "29.4.0",
 		"tslib": "2.6.2",
-		"typescript": "5.5.2",
+		"typescript": "5.1.3",
```
oliverabrahams, line 48: This makes me a bit concerned. Would it be straightforward to update flexible-frontend to v5.5.2? 
We've had issues in the past with mismatched versions of shared dependencies. Having two different versions of TypeScript and no separate lockfiles for each package can lead to dependency resolution problems, where the wrong version is used unexpectedly.

andrewHEguardian:
To avoid some of the issues with dependencies being out of date in consuming apps, and to gain the benefits of colocating the code we've decided not to publish this component from this repo

PR #2365, opened 2026-05-05
add css snippet to restore styling for html tags in crossword clues

andrew-nowak:
## What are you changing?

Add a little css snippet to re-add basic HTML styling for tags used to markup crossword clues. I believe the basic HTML styles are overridden by a CSS reset.

## Why?

Formatted clues appear with the intended formatting


oliverabrahams (approved):
Looks good

PR #2386, opened 2026-05-18
Upgrade TypeScript to 6.0.3

coldlink:
Upgrades the monorepo from TypeScript 5.9.3 to 6.0.3, in preperation for Typescript 7. References the [official TS6 migration guide](https://aka.ms/ts6) and [privatenumber's TS6 upgrade gist](https://gist.github.com/privatenumber/3d2e80da28f84ee30b77d53e1693378f), with some help from Copilot + Clause Sonnet 4.6 to help fix errors.

- Bumped `typescript` to `6.0.3` in the root and all packages that declare it as a peer dependency.
- Removed `"baseUrl": "."` from tsconfig.base.json which is deprecated in TS6.

**Project changes**

In TS6, `compilerOptions.types` defaults to an empty array when `typeRoots` is unset, meaning ambient type packages are no longer auto-discovered. Added explicit `types` entries to each package that relies on them:
- `"types": ["jest"]`
  -  all packages with Jest tests
- `"types": ["jest", "chai"]`
  -  `react-crossword` (also uses `@vitest/expect` which references the `Chai` namespace)
  - `source-development-kitchen`
- `"types": ["node"]`
  -  `browserslist-config`, `newsletter-types`, `prettier`, scripts (their `tsconfig` includes `eslint.config.mjs` which transitively imports Node.js types)

`@guardian/libs` - Refactored the exported anonymous class expression to a named class declaration (`StorageManager`) to fix **TS4094** ("Property of exported anonymous class type may not be private"). TS6 enforces this for private fields (`#field`) on anonymous classes.

`@guardian/core-web-vitals` - FID (First Input Delay) is no longer a Core Web Vital and was [removed](https://github.com/GoogleChrome/web-vitals/blob/main/docs/upgrading-to-v5.md) from `web-vitals` in v5. Upgraded `web-vitals@4` to `5.2.0` and removed all FID references:
- Removed `onFID`, `FIDMetricWithAttribution` from index.ts
- Removed `fid` field from `CoreWebVitalsPayload`
- Updated `LCPAttribution.element` → `.target` (renamed in v5)
- Removed `INPAttribution.interactionTargetElement` (removed in v5)
- Updated tests accordingly

**Overrides**

- `@package-json/types`: overridden to `0.0.13` as v0.0.12 has a **TS2411** bug where optional properties (`node?: string`) inside objects with `[k: string]: string` index signatures are incompatible under strict mode. This surfaced because package `tsconfig.json` files include `eslint.config.*` via `"include": ["**/*"]`, which transitively imports `eslint-plugin-import-x` → `@package-json/types`.

---
  - https://app.asana.com/0/0/1214972512316727

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
 		"prettier": "3.8.0",
 		"sort-package-json": "3.6.0",
 		"syncpack": "13.0.0",
-		"typescript": "5.9.3",
+		"ts-jest": "29.4.11",
+		"typescript": "6.0.3",
 		"wireit": "0.14.12"
 	},
 	"packageManager": "pnpm@10.28.1+sha512.7d7dbbca9e99447b7c3bf7a73286afaaf6be99251eb9498baefa7d406892f67b879adb3a1d7e687fc4ccc1a388c7175fbaae567a26ab44d1067b54fcb0d6a316",
@@ -41,6 +42,7 @@
 			]
 		},
 		"overrides": {
+			"@package-json/types": "0.0.13",
 			"effect": "~3.20.0",
 			"lodash": "~4.18.0",
 			"minimatch": "~10.2.3"
--- a/.changeset/frank-pillows-go.md
+++ b/.changeset/frank-pillows-go.md
@@ -1,0 +1,22 @@
+---
+'@guardian/source-development-kitchen': major
+'@guardian/identity-auth-frontend': major
+'@guardian/browserslist-config': major
+'@guardian/newsletter-types': major
+'@guardian/consent-manager': major
+'@guardian/core-web-vitals': major
+'@guardian/react-crossword': major
+'@guardian/eslint-config': major
+'@guardian/identity-auth': major
+'@guardian/ab-react': major
+'@guardian/prettier': major
+'@guardian/ab-core': major
+'@guardian/source': major
+'@guardian/libs': major
+'@guardian/tsconfig': major
+---
+
+Update Typescript to v6 (6.0.3):
+
+- Use https://gist.github.com/privatenumber/3d2e80da28f84ee30b77d53e1693378f as a reference to update your own codebase to TS6.
+- If you use `ts-jest` to run jest tests, update to v29.4.11 minimum to ensure compatibility with TS6.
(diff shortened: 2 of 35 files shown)


libs/@guardian/tsconfig/tsconfig.json
```diff
@@ -1,14 +1,11 @@
...
 		"module": "esnext",
 		"moduleResolution": "bundler",
 		"noImplicitReturns": true,
 		"noUncheckedIndexedAccess": true,
 		"noUnusedLocals": true,
 		"resolveJsonModule": true,
-		"strict": true,
 		"target": "esnext"
 	}
 }
```
coldlink, line 14: The things that have been removed are now set to those values by default, so no longer need to set them in this file.

patches/ts-jest@29.4.9.patch
```diff
@@ -0,0 +1,78 @@
+diff --git a/dist/legacy/compiler/ts-compiler.js b/dist/legacy/compiler/ts-compiler.js
```
coldlink: Backported from https://github.com/kulshekhar/ts-jest/pull/5280, can remove this patch once a new version of ts-jest has been published!

package.json
```diff
@@ -41,13 +41,15 @@
 			]
 		},
 		"overrides": {
+			"@package-json/types": "0.0.13",
```
coldlink, line 45: overridden to `0.0.13` as v0.0.12 has a **TS2411** bug where optional properties (`node?: string`) inside objects with `[k: string]: string` index signatures are incompatible under strict mode. This surfaced because package `tsconfig.json` files include `eslint.config.*` via `"include": ["**/*"]`, which transitively imports `eslint-plugin-import-x` → `@package-json/types`

coldlink:
Blocked by https://github.com/guardian/csnx/pull/2397

patches/ts-jest@29.4.9.patch
coldlink: ESM changes now in https://github.com/guardian/csnx/pull/2397

oliverabrahams (approved):
Looks good! Nice work

PR #2397, opened 2026-05-19
Enable Jest/ts-jest ESM support in preperation for typescript v6+

coldlink:
## What are you changing?

- Enable [Jest ESM](https://jestjs.io/docs/ecmascript-modules) and [ts-jest ESM](https://kulshekhar.github.io/ts-jest/docs/guides/esm-support) support
- Update base jest configuration to use ESM preset
- Install `@jest/globals` for `jest` variable support
- Go through each package one-by-one and fix tests/types/linting as required to enable new support e.g. by using dymanic imports in order to support mocking

## Why?

- Typescript 6 has deprecated `moduleResolution=node10`, but ts-jest in it's default state applies this regardless of what is in your tsconfig.json
- In [ts-jest 29.4.10](https://github.com/kulshekhar/ts-jest/blob/main/CHANGELOG.md#29410-2026-05-18) a bug was fixed that "respect tsconfig moduleResolution instead of forcing Node10 ([1bffffc](https://github.com/kulshekhar/ts-jest/commit/1bffffc667557c173ae0c1f93dd436920775dac4))"
- However jest by default runs in commonjs mode, and ts-jest will continue to run in `moduleResolution=node10` mode as long as jest is doing this, despire CSNX setup to output ESM by default, with a comment in the mentioned commit confirming this
- So we need to first enable ESM support in Jest, so that ts-jest respects `moduleResolution` in our tsconfig file, and then that means we can upgrade to Typescript 6+
- We also now respect modern standards

## Notes

Maybe Jest isn't the best long term tool for JS/TS testing? Could we look into alternatives, such as [vitest](https://vitest.dev/)?

---
  - https://app.asana.com/0/0/1214972512316724

### DIFF
--- a/configs/jest.config.js
+++ b/configs/jest.config.js
@@ -1,3 +1,7 @@
+import { createDefaultEsmPreset } from 'ts-jest';
+
+const presetConfig = createDefaultEsmPreset();
+
 /** @typedef {import("jest").Config} Config  */
 export const config = {
 	clearMocks: true,
@@ -6,8 +10,6 @@ export const config = {
 	transformIgnorePatterns: [
 		'node_modules/.pnpm/(?!@guardian|use-local-storage-state)',
 	],
-	transform: {
-		'^.+\\.[tj]sx?$': ['ts-jest'],
-	},
 	coveragePathIgnorePatterns: ['/__generated__/'],
+	...presetConfig,
 };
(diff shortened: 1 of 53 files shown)


coldlink:
No longer needed with ts-jest v29.4.11 https://github.com/kulshekhar/ts-jest/blob/main/CHANGELOG.md#29411-2026-05-21

PR #2398, opened 2026-05-19
Improve Popover component

cemms1:
## What are you changing?

- Use  `trigger` instead of `anchorElement` for the name of button which controls the opening of the popover
- Use `content` instead of `children` for the popover content itself
- Updates accessibility attributes

## Why?

- Standardising the naming and structure of the popover component to be closer to other industry examples
- Improving accessibility of the popover component

### DIFF
--- a/libs/@guardian/source-development-kitchen/src/react-components/popover/Popover.tsx
+++ b/libs/@guardian/source-development-kitchen/src/react-components/popover/Popover.tsx
@@ -190,7 +190,7 @@ export const Popover = ({
 				]}
 				role="dialog"
 				aria-modal={false}
-				aria-labelledby={titleId}
+				aria-labelledby={hideTitle ? title : titleId}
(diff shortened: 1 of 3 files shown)


libs/@guardian/source-development-kitchen/src/react-components/popover/Popover.tsx
```diff
@@ -187,16 +189,23 @@ export const Popover = ({
 					!!cssOverrides && cssOverrides,
 				]}
 				role="dialog"
-				aria-label={title}
+				aria-modal={false}
+				aria-labelledby={titleId}
```
oliverabrahams: if hideTitle = true then the titleId component wont be there so could set this to 
```suggestion
				aria-labelledby={hideTitle ? title : titleId}
```

oliverabrahams (approved):
Nice, looks good!

PR #2418, opened 2026-06-23
Update tabbing behaviour for `Popover` component

cemms1:
## What are you changing?

Updates the tab behaviour for the Popover component in `source-development-kitchen` so that when the popover is opened, the **keyboard focus is trapped** within the popover until the user takes an action to dismiss it 

The popover can still be dismissed by using the escape key or by clicking the dismiss/close button 

### Screen recordings
**Before**

https://github.com/user-attachments/assets/1e04365c-96e9-426e-98cf-49ee50e59708

**After**

https://github.com/user-attachments/assets/937b52a7-1dac-40ec-a4fc-1a914d6ab1da

## Why?

After discussing as a team, it felt more intuitive for keyboard users and those using screen readers for the focus to be trapped when opening the Popover since otherwise the focus moves to the next tabbable component on the page, which can be _behind_ the popover in terms of z-index value. Being able to focus on elements which are not entirely visible on the screen feels like poor UX.

The guidance for this online is mixed and in general suggests that you should only trap focus for _"modal"_ popover content and not for non-modal popovers. This is not technically a modal popover


libs/@guardian/source-development-kitchen/src/react-components/popover/Popover.tsx
```diff
@@ -149,21 +150,55 @@ export const Popover = ({
...
+				event.preventDefault();
+				lastTabbableItem?.focus();
+			}
+			// Tab moves focus from last element to first element
+			else if (document.activeElement === lastTabbableItem) {
+				// Move focus to first tabbable element
+				event.preventDefault();
+				firstTabbableItem?.focus();
+			}
+		};
```
dskamiotis, line 174: Nice implementation. I found this really easy to read—the variable names and comments make the focus-trapping logic really clear at a glance. 👍

dskamiotis (approved):
This looks good to me, nice work 🎆

deedeeh (approved):
Much better UX and accessible @cemms1 🎉

jamesmockett (approved):
Nice improvement and beautifully commented ✨

PR #2428, opened 2026-07-15
Add Feast API scopes for "Saved from web" endpoints

andresilva-guardian:
## What are you changing?

- We added two new OAuth scopes to the `scopes` array in `getIdentityAuth()` in index.ts, `guardian.feast-api.saved-from-web.update.self` and `guardian.feast-api.saved-from-web.read.self`, each with an inline comment explaining what it grants
- We kept every existing scope in place, unchanged and in the same order
- We added a changeset, moody-seas-sing.md, marking this as a minor version bump for `@guardian/identity-auth-frontend`
- We ran the package's existing test suite to confirm nothing else broke

## Why?

- Dotcom-rendering, and any other consumer of `@guardian/identity-auth-frontend`, needs to request access tokens that are valid for the new "Saved from web" endpoints on the Feast API, added in personalised-app-experience#256
- `PUT /v2/saved-from-web/{recipeId}` requires the `guardian.feast-api.saved-from-web.update.self` scope, and `GET /v2/saved-from-web/{recipeId}` requires the `guardian.feast-api.saved-from-web.read.self` scope
- These scopes have already been added to the "theguardian" Okta OAuth client in the identity-platform repo's Okta Terraform, so Okta will allow them to be requested, and consent is implicit, so we don't need any consent screen or UI changes
- We are keeping these scopes in the single global `scopes` array, consistent with how every other scope in this list is handled today, since this package doesn't support per-feature or per-call scoping


AshCorr:
> We ran the package's existing test suite to confirm nothing else broke

Not entirely sure the packages test suite is broad enough to verify the interaction between DCR <-> @guardian/identity-auth <-> Gateway

Might be worth doing a pre-release by adding the "🐥 Canaries" label to this PR, so you can test this change in DCR CODE? I'm not entirely sure how @guardian/identity-auth behaves when a new scope is added, it shouldn't break, but I'm slightly worried about it signing users out who are missing the scope resulting in everyone getting signed out.

juabara (approved):

akinsola-guardian (approved):

PR #2429, opened 2026-07-16
Setup AB Test logic to pass new targetingParam to Sourcepoint depending on ab test and state/region cookie

akinsola-guardian:
## What are you changing?

- Adds AB-test logic by checking the `gu_client_ab_tests` cookie to determine which AB test group the user belongs to, and the `GU_geo_country_region` cookie to determine if the user's state is part of the list defined in `AB_TEST_US_STATES`. 
- These two cookies are used to determine whether the `abTestGroup` value is passed as a targetingParam to Sourcepoint.
- This test only applies to users in the US in the states defined in `AB_TEST_US_STATES`

This will have no effect until the Sourcepoint Campaign changes to receive new targeting param and the AB test cookie is set up in DCR.

## Why?

- Implementing AB Test for US states.


libs/@guardian/consent-manager/src/lib/sourcepointConfig.ts
```diff
@@ -7,8 +7,9 @@ export const PROPERTY_ID_MAIN = 7417;
 export const PROPERTY_ID_SUBDOMAIN = 38161;
 export const PROPERTY_ID_AUSTRALIA = 13348;
 
+export const PROPERTY_HREF_AUS = 'https://au.theguardian.com';
 export const PROPERTY_HREF_SUBDOMAIN = 'https://subdomain.theguardian.com';
-export const PROPERTY_HREF_MAIN = 'https://test.theguardian.com';
+export const PROPERTY_HREF_MAIN = 'https://www.theguardian.com';
```
akinsola-guardian, line 12: The propertyHref is `*.theguardian.com`

libs/@guardian/consent-manager/src/usAbtest.test.js
```diff
@@ -0,0 +1,147 @@
...
+		it('returns undefined when the ab test cookie is null', () => {
+			mockGetCookie.mockReturnValue(null);
+			expect(getUsAbTestGroup()).toBeUndefined();
+		});
+	});
+
+	describe('isUserInAbTest', () => {
+		it('returns true when the user is in a supported US state and is in the control ab test group', () => {
+			mockGetCookie.mockImplementation(({ name }) => {
+				if (name === guCountryRegionCookieName) {
```
brunocatarinoguardian, line 79: Maybe I'm misunderstanding the code, but do we need a conditional mock implementation if there's only one thing being tested? I'd understand if the mock was shared by different tests, but just used on this one, not sure I get it?

brunocatarinoguardian: ah nevermind, it's because you always check the two cookies and you have no other safe way to specify which specific cookie you're mocking? Ignore me!

brunocatarinoguardian (approved):
Don't know the repo, so take it with a grain of salt, but overall looks good.

PR #2455, opened 2026-08-13
Temporarily prevent publishing of Pulse package

jamesmockett:
## What are you changing?

Add Pulse package to Changesets ignore list to temporarily prevent publishing[^1]

## Why?

We haven't yet set up trusted publishing for this package so the publishing workflow fails with an error. This package is also empty at the moment so there is nothing of value to publish.

[^1]: https://changesets.dev/guide/config#ignore

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -8,7 +8,13 @@
 	"baseBranch": "main",
 	"updateInternalDependencies": "patch",
 	"bumpVersionsWithWorkspaceProtocolOnly": false,
-	"ignore": ["github-pages", "@configs/*", "coverage", "storybooks"],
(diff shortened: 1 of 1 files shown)


coldlink:
was literally just looking at this!

I think the best was is actually to ignore it in `.changeset/config.json`

```
"ignore": [
  "github-pages",
  "@configs/*",
  "coverage",
  "storybooks",
  "@guardian/pulse"
]
```

This excludes it from both release-versioning and the `pnpm changeset publish` commands used by `npm-publish.yml`. 
And we can remove it easily when we want to be able to publish again.

coldlink (approved):

PR #2464, opened 2026-08-20
ADR | Document how Pulse is made available to consumers

coldlink:
Adds an ADR covering the decision to expose Pulse tokens as CSS custom properties with data-attribute theme contexts.

It records the options considered, Emotion usage, nested-theme inheritance caveats, API-stability implications, and the potential for a future provider API.

---
  - https://app.asana.com/0/0/1217369629582006


jamesmockett (approved):
This is great and the 'API stability' section covers the case that we _may_ need to refine things once we start using it more broadly ✨

PR #2489, opened 2026-09-11
Rollout AB Test only to South Carolina intially

akinsola-guardian:
## What are you changing?

- This PR comments out other US states for the US AB banner test.

## Why?

- This will be rolled out the one state for testing before rolling out to other 11 states.

### DIFF
--- a/libs/@guardian/consent-manager/src/usAbtest.test.js
+++ b/libs/@guardian/consent-manager/src/usAbtest.test.js
@@ -18,26 +18,27 @@ describe('usAbtest', () => {
 	const identityAndTrustConsentAbTestCookieName =
 		'identity-and-trust-consent-rr-banner-us';
 	const abTestCookieName = 'gu_client_ab_tests';
+	const usStateInSupportedList = 'SC';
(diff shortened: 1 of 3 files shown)


libs/@guardian/consent-manager/src/usAbtest.test.js
```diff
@@ -18,26 +18,27 @@ describe('usAbtest', () => {
 	const identityAndTrustConsentAbTestCookieName =
 		'identity-and-trust-consent-rr-banner-us';
 	const abTestCookieName = 'gu_client_ab_tests';
+	const usStateInSupportedList = 'SC';
```
ioannakok: Can you perhaps use `AB_TEST_US_STATES` list from `usAbtest.ts` instead so that you don't need to update the list in two places?

ioannakok (approved):

libs/@guardian/consent-manager/src/usAbtest.test.js
akinsola-guardian: Thank you. Just did this.

